PR #12, opened 2021-09-08
11-Fix Maven configuration

atextor:
Adds missing plugin versions and sets versions in plugin configuration


bs-jokri (approved):
@atextor great! Thanks to spot that! LGTM

PR #16, opened 2021-09-15
Add configuration for release workflow

LuLeRoemer:
This PR contains the necessary configuration for the GitHub Actions release workflow to release the SDS SDK to maven and GitHub.

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -1,0 +1,83 @@
+name: Release SDS SDK
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: 'Version number of the release'
+        required: true
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.1
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,83 @@
...
+          server-password: OSSRH_TOKEN
+          gpg-private-key: ${{ secrets.PGP_KEY }}
+          gpg-passphrase: PGP_KEY_PASSWORD
+      - name: Setup Git
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: Sanity check version
+        run: |
+          current_version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```
eriksven, line 33: Is the current version still required here? It seems like it is not used.

eriksven: We can fix this with a later PR, when we make further modifications to the release workflow. E.g., we should add the CLI to the files to build and be part of the GitHub release.

eriksven (approved):
Besides the comment above this looks good to me.

PR #20, opened 2021-09-28
change mvn command in pr workflow

eriksven:
Changes mvn command in pr workflow from test to install.


eriksven:
base branch for PR is not correct.

PR #47, opened 2021-11-05
Reintroduce issue templates to repository

bs-jokri:


eriksven (approved):
LGTM and is similar to the approach for the SDS BAMM repository

PR #50, opened 2021-11-05
Improves documentation about CLI usage

bs-jokri:
- describes that folder structure is required for aspect models
- updated options according to version 1.0.0 of cli


eriksven (approved):
thank you for updating the CLI Readme

PR #80, opened 2021-12-16
Include hashCode in generated POJOs

### LINKED ISSUE #35: Include hashcode in generated POJOs
**Description**

The domain classes which are generated from an Aspect model currently only contain an implementation of the `equals` method. Additionally the generated code should contain an implementation for the `hashcode` method.

georgschmidtdumont:
In addition to the equals method, the generated Java classes now contain
an implementation for the hashCode method.

fixes #35

### DIFF
--- a/core/sds-aspect-model-java-generator/src/main/resources/java-pojo-hashcode-method-lib.vm
+++ b/core/sds-aspect-model-java-generator/src/main/resources/java-pojo-hashcode-method-lib.vm
@@ -2,6 +2,12 @@
 @Override
 public int hashCode() {
     return Objects.hash(
+    #if ( $element.isComplex() )
+        #set( $complexElement = $ComplexType.cast( $element ) )
+        #if ( $complexElement.getExtends().isPresent() )
+            super.hashCode(),
+        #end
+    #end
     #foreach( $property in $element.getProperties() )
         ${property.getPayloadName()}
         #if( $foreach.hasNext) , #end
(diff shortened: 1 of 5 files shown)


core/sds-aspect-model-java-generator/src/main/resources/java-pojo-hashcode-method-lib.vm
```diff
@@ -0,0 +1,11 @@
+#macro( javaPojoHashCodeMethod )
+@Override
+public int hashCode() {
+    return Objects.hash(
+    #foreach( $property in $element.getProperties() )
```
MelleD, line 5: Missing super.hashCode call

georgschmidtdumont: The call to `super.equals` is also missing. I will update the logic for both methods to have consistent `equals` and `hashcode` implementations.

atextor: The classes are never generated with any explicit superclass, so what is the point?

georgschmidtdumont: In case an Entity extends another Entity, then the corresponding class is generated with a superclass. Lets say we the `Car` Entity which extends the `Vehicle` Entity. Then the class fro the `Car` Entity is generated as
`public class Car extends Vehicle`.

atextor: You're right, makes sense!

atextor (approved):

PR #107, opened 2022-02-02
Generate valid code for Aspects without Properties

### LINKED ISSUE #103: [BUG] Invalid code generated for Aspects without Properties

atextor:
This PR fixes POJO and static classes generation for Aspects without Properties.
Furthermore it introduces systematic testing of code generation buy running the
code generation on all test models; all errors in code generation preventing to
run theses systematic tests successfully are fixed as well.

Fixes #103

### DIFF
--- a/core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
+++ b/core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
@@ -116,25 +116,18 @@
     #if( $optionalValue.isEmpty() )
         Optional.empty()
     #else
-        #set( $objectClass = $optionalValue.get().getClass() )
-        $codeGenerationConfig.getImportTracker().importExplicit( $objectClass )
+        $codeGenerationConfig.getImportTracker().importExplicit( $util.getDataTypeClass( $dataType ) )
         #set( $xsdType = $ResourceFactory.createResource( $dataType.getUrn() ) )
         #set( $valueExpression = $optionalValue.get().toString() )
         #if( $dataType.getUrn().endsWith( "#float" ) )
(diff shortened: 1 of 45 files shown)


DvSmi (approved):

core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
```diff
@@ -337,7 +335,7 @@
           $util.toConstant( $property.getName() ) = new StaticConstraintProperty<$propertyType, #getConstraintClassName(), #getCharacteristicClassName()>(
     #end
 #else
-    #if( $util.hasContainerType( $property ) )
+    #if( $util.hasContainerType( $property ) && !$propertyType.startsWith( "Map" ) )
```
georgschmidtdumont, line 338: Why are you excluding `Map` here?

georgschmidtdumont (changes requested):

core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
atextor: This is a fix to align this section with the generation of the `getContainedType` method starting in line 392. Map is (correctly) marked as a "container type", which is required for example in the generation of the `getPropertyType` method, which in this case returns a `Class<Map>`. The reason Map is excluded here is because maps are generated for Properties such as `bamm:description`, which are represented as a `Map<Locale, String>` in the static class, but for which no `StaticContainerProperty`s should be generated but rather `StaticProperty`s as the are no containers like collections.

georgschmidtdumont (approved):

PR #116, opened 2022-03-03
#100: Restructure bamm-cli interface

### LINKED ISSUE #100: [Task] Clean up bamm-cli interface
All of the bamm-cli functionality is currently controlled using a flat list of switches, some of which apply only to one or few actual functions. A best practice for CLIs here is to work with subcommands (cf. eg. git, az cli, kubectl, ...). This should be cleaned up. Here it would make sense to switch from the currently used JCommander to [picocli](https://picocli.info/).

A clean refactoring/reimplementation of the bamm-cli using picocli; structured with subcommands that make sense, each of which has its own set of switches.

RaMisess:
Fixes #100


tools/bamm-cli/src/main/java/io/openmanufacturing/sds/BammCli.java
```diff
@@ -12,542 +12,52 @@
...
-import io.vavr.control.Option;
-import io.vavr.control.Try;
-
+import org.fusesource.jansi.AnsiConsole;
+
+import io.openmanufacturing.sds.aspect.AspectCommand;
+import picocli.CommandLine;
+
+@CommandLine.Command( name = BammCli.COMMAND_NAME,
+      description = "Command line tool for working with Aspect Models and the Digital Twin System",
```
atextor, line 21: Remove the reference to Digital Twin System

tools/bamm-cli/src/main/java/io/openmanufacturing/sds/aspect/to/AspectToOpenapiCommand.java
```diff
@@ -0,0 +1,174 @@
...
+   private String outputFilePath = "-";
+
+   @CommandLine.ParentCommand
+   private AspectToCommand parentCommand;
+
+   @Override
+   public void run() {
+      final AspectModelOpenApiGenerator generator = new AspectModelOpenApiGenerator();
+      final Aspect aspect = AspectModelLoader.fromVersionedModelUnchecked( loadModelOrFail( parentCommand.parentCommand.getInput() ) );
+      if ( generateJsonOpenApiSpec ) {
```
atextor, line 101: Currently, either `-j` or `-y` must be given, otherwise the command does nothing. I would either output an error message when both are not given or maybe even better, use one of them as default. I think most people would prefer YAML over JSON for OpenAPI, so I'd change the command so that when no option is given, YAML is generated, and if `-j` is given, use JSON instead of YAML.The `-y` switch can then be removed completely.

tools/bamm-cli/src/main/java/io/openmanufacturing/sds/aspect/to/AspectToSchemaCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+import io.openmanufacturing.sds.AbstractCommand;
+import io.openmanufacturing.sds.aspect.AspectToCommand;
+import io.openmanufacturing.sds.aspectmodel.generator.jsonschema.AspectModelJsonSchemaGenerator;
+import io.openmanufacturing.sds.exeption.CommandException;
+import io.openmanufacturing.sds.metamodel.Aspect;
+import io.openmanufacturing.sds.metamodel.loader.AspectModelLoader;
+import picocli.CommandLine;
+
+@CommandLine.Command( name = AspectToSchemaCommand.COMMAND_NAME,
```
atextor, line 29: Maybe this should be renamed to `AspectToJsonSchemaCommand` since in theory we could also write other "schema"s, such das RDF Schema or XML Schema.

tools/bamm-cli/src/test/java/io/openmanufacturing/sds/ApplicationIntegrationTest.java
```diff
@@ -72,28 +83,23 @@ public void beforeEach() throws IOException {
...
    @Test
    public void testNoArgs() throws Exception {
-      assertThat( catchSystemExit( bammCli::run ) ).isEqualTo( 1 );
+      assertThat( catchSystemExit( () -> bammCli.run( new String[0] ) ) ).isEqualTo( 0 );
    }
 
    @Test
    public void testWrongArgs() throws Exception {
-      assertThat( catchSystemExit( () -> bammCli.run( "-i", "notexists" ) ) ).isEqualTo( 1 );
+      assertThat( catchSystemExit( () -> bammCli.run( "-i", "notexists" ) ) ).isEqualTo( 2 );
```
atextor, line 96: Why is the expected error code 2?
I would expect one of the following behaviours:
1. Successful run returns 0, failure returns 1
2. Successful run returns 0, failure returns some error code, but all error possible codes are documented

atextor (changes requested):

atextor:
It seems there is currently no way to display the help for sub-subcommands, e.g., the follwing works:
```sh
bamm-cli help
bamm-cli help aspect
bamm-cli aspect --help
bamm-cli aspect help to
```
but this doesn't, i.e., does not show what I would expect:
```sh
bamm-cli aspect to help openapi  # should show the help for the openapi subcommand
bamm-cli aspect to openapi --help  # should show the help for the openapi subcommand
bamm-cli aspect to --help  # should show the help for the 'to' subcommand
```

atextor:
Can you please rebase the PR against main? This would make review much easier.

PR #136, opened 2022-03-25
Make presence of `bamm:properties` and `bamm:operations` optional

atextor:
The current codebase is already able to handle the absence of the
`bamm:properties` and `bamm:operations` attributes, even though this was not
allowed until recently. With
OpenManufacturingPlatform/sds-bamm-aspect-meta-model#105, this becomes valid.
This PR adds tests that verify that this is handled correctly in the sds-sdk.


core/sds-aspect-model-serializer/src/test/java/io/openmanufacturing/sds/aspectmodel/serializer/RdfModelCreatorVisitorTest.java
```diff
@@ -42,13 +42,18 @@
     * Exclude the test models that contain unused elements; the resulting serialized models can
     * never be identical to the original because the unused elements are ignored when loading the
     * model and can therefore not be serialized.
+    * Furthermore, exclude the test model of an Aspect without bamm:properties and bamm:operations,
```
georgschmidtdumont, line 45: Shouldn't we update the serialization instead?

atextor: If it's valid to write:
```turtle
:Foo a bamm:Aspect .
```
as well as:
```turtle
:Foo a bamm:Aspect ;
  bamm:properties () .
```
and you load the model into the Java representation, you will end up with an empty `List<Property>` either way.
Now, the serializer will write a model that looks like the second of the two examples (with `bamm:properties`), so in the test for the first example there would be a mismatch between source model and serialization result.
If we'd change the serializer to write a model that looks like the first of the two examples (without `bamm:properties`), you would have the same problem the other way round: All input models _with_ `bamm:properties` would look different than their serialization.

georgschmidtdumont (approved):

PR #156, opened 2022-05-31
Support Abstract Properties

### LINKED ISSUE #125: [Task] Support `bamm:refines` replacement
In OpenManufacturingPlatform/sds-bamm-aspect-meta-model#34, the `bamm:extends` keyword and the concept of Abstract Entities was introduced to replace the complicated `bamm:refines` construct. Since the concept of Abstract Entities alone is not sufficient to replace `bamm:refines`, OpenManufacturingPlatform/sds-bamm-aspect-meta-model#110 proposes alternative additional modeling techniques. Whatever is used as a replacement needs to be supported in the SDK.

Implement loading/handling the corresponding model changes

atextor:
Fixes #125. This implements the `bamm:AbstractProperty` and proper corresponding
semantics of `bamm:extends`.

### DIFF
--- a/core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
+++ b/core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
@@ -26,6 +26,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.RDFNode;
 import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.rdf.model.Statement;
 import org.apache.jena.vocabulary.RDF;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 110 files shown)


core/sds-aspect-model-document-generators/src/main/java/io/openmanufacturing/sds/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
```diff
@@ -239,18 +242,38 @@ public JsonNode visitHasProperties( final HasProperties element, final ObjectNod
       return context;
    }
 
+   private Characteristic determineCharacteristic( final Property property ) {
+      if ( property.getCharacteristic() != null ) {
```
georgschmidtdumont, line 246: In which scenario is the Characteristic `null`?

core/sds-aspect-model-java-generator/src/test-shared/java/io/openmanufacturing/sds/test/shared/compiler/JavaCompiler.java
```diff
@@ -45,19 +56,45 @@ public static Map<QualifiedName, Class<?>> compile( final List<QualifiedName> lo
             .map( key -> new CompilerInput( key.toString(), sources.get( key ) ) )
             .collect( Collectors.toList() );
 
+      if ( System.getProperty( WRITE_SOURCES_PROPERTY ) != null ) {
```
georgschmidtdumont, line 59: Why did you make this optional? To me it sound really helpful to have the generated classes from the tests written to an explicit directory.

core/sds-aspect-model-java-generator/src/test-shared/java/io/openmanufacturing/sds/test/shared/compiler/JavaCompiler.java
```diff
@@ -45,19 +56,45 @@ public static Map<QualifiedName, Class<?>> compile( final List<QualifiedName> lo
...
       };
 
       final List<String> compilerOptions = List.of( "-classpath", System.getProperty( "java.class.path" ) );
-      compiler.getTask( null, manager, diagnosticListener, compilerOptions, null, compilerInput ).call();
+      compiler.getTask( null, manager, diagnosticListener, compilerOptions, null, compilerInput ).
+
+            call();
 
       return loadOrder.stream()
-            .collect( Collectors.toMap( Function.identity(), qualifiedName -> defineAndLoad( qualifiedName, manager ) ) );
```
georgschmidtdumont, line 60: Please reformat.

georgschmidtdumont (changes requested):

core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
```diff
@@ -145,56 +143,80 @@ public static MetaModelBaseAttributes from( final KnownVersion metaModelVersion,
...
-            .substring( 0, 7 );
-      return ImmutableList.copyOf( model.listStatements( modelElement, RDF.type, (RDFNode) null ) ).stream()
-            .findFirst()
-            .filter( statement -> statement.getObject().isResource() )
-            .map( statement -> statement.getObject().asResource() )
-            .map( type -> metaModelResourceResolver.getAspectModelUrn( type.getURI() ) )
-            .flatMap( Value::toJavaOptional )
-            .map( AspectModelUrn::getName )
-            .orElse( "" ) + randomPart;
+   private static Optional<String> getName( final Resource modelElement, final BAMM bamm ) {
```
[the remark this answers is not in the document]
atextor: They were in fact the same method, one of them was not used. I suppose this was either an error during refactoring or merging. The unused (getNameFromInheritanceTree) is removed now.

core/sds-aspect-model-document-generators/src/main/java/io/openmanufacturing/sds/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
atextor: The Characteristic can be `null` if the Property is abstract.

core/sds-aspect-model-java-generator/src/test-shared/java/io/openmanufacturing/sds/test/shared/compiler/JavaCompiler.java
atextor: Yes, I would like to have this as well. The problem is that many model test models define model elements with the same name (e.g. "TestEntity") but with different structure and semantics. When you selectively run one test, this is no problem. When you run the test suite, where code for all test models is generated, you end up in a situation where one test overwrites previously generated code from another test.
Let's say AspectA has a TestEntity and AspectB also has TestEntity but with a different structure. Then AspectB's TestEntity's constructor does not match the `new TestEntity` cann in AspectA and you end up with compilation errors. We would have to cleanly move each test model into its own namespace. I consider this out of scope for this PR though.

core/sds-aspect-model-document-generators/src/main/java/io/openmanufacturing/sds/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
georgschmidtdumont: Would it not make sense to define the Characteristic as optional in the `Property` class?

atextor: You're correct. I have now updated `getCharacteristic()` in Property to return `Optional<Characteristic>`. This communicates in a clearer way that there are cases in which no Characteristic is present.

georgschmidtdumont (approved):

Issue #184, opened 2022-07-27
[BUG] Entity without property generates non compilable java code


DvSmi:
**Describe the bug**
Defining an Aspect with an Entity without any further Properties, and the Entity extending an AbstractEntity with a property, leads to an incorrect generation of the hashCode method.
```
@Override
public int hashCode() {
    return Objects.hash(
            super.hashCode(),
    );
```

atextor:
Fixed by #185

PR #185, opened 2022-07-27
Fixes bug in java code generation in case an entity has no property

DvSmi:
solves #184 
Fixes bug in java code generation in case an entity has no property, but extends another entity.

This leaded to an error in the hashCode generation.


core/sds-aspect-model-java-generator/src/main/java/io/openmanufacturing/sds/aspectmodel/java/AspectModelJavaUtil.java
```diff
@@ -535,11 +535,15 @@ public static String objectEqualsExpression( final StructureElement element ) {
             .collect( Collectors.joining( " && " ) );
    }
 
-   public static String objectsHashCodeExpression( final StructureElement element ) {
-      return element.getProperties().stream()
+   public static String objectsHashCodeExpression( final StructureElement element, final boolean needsLeadingComma ) {
```
atextor, line 538: Can the logic of adding the leading comma be moved into the template rather than the method?

atextor (changes requested):

core/sds-aspect-model-java-generator/src/main/java/io/openmanufacturing/sds/aspectmodel/java/AspectModelJavaUtil.java
DvSmi: I couldn't find a nice solution as we need to know if there are any elements and only then we need the comma. So we would partially the same logic of finding the elements in the template which seems not much better than.
Or do you see a better way for that?

atextor: Yes, you would have to have to build another nested `#if`, but I admit that this is not pretty either. So I guess we leave it like that.

atextor (approved):

PR #205, opened 2022-08-10
Fix model resolution

### LINKED ISSUE #202: Fix model resolution where rdf lists / anonymous nodes are involved
**What**
When a ttl with an Aspect Model refers an element defined in another ttl locally on storage, 
the validation can fail with a message like "Error on element MyChar: Property may only have 1 value, but found 2".
As an example see primary aspect and secondary aspect from the attached zip.
[io.openmanufacturing.test.zip](https://github.com/OpenManufacturingPlatform/sds-sdk/files/9291502/io.openmanufacturing.test.zip)
In this example, the Characteristic TrafficLight of class enumeration with a list of values is affected.
If TrafficLight is changed to be a base Characteristic with dataType string, the issue does not occur.

**Where**
This can be reproduced with both the Aspect Model Editor (4.0.0 or 4.0.1) as well as the bamm cli.

**Screenshots**
image

**Additional context**

atextor:
Fixes #202


core/sds-aspect-model-resolver/src/main/java/io/openmanufacturing/sds/aspectmodel/resolver/AspectModelResolver.java
```diff
@@ -130,85 +112,134 @@ public Try<VersionedModel> resolveAspectModel( final ResolutionStrategy resolver
...
    }
 
-   private Model addModelWithoutOverwritingEmptyPrefix( final Model target, final Model modelToAdd ) {
-      if ( !target.getNsPrefixMap().containsKey( "" ) && modelToAdd.getNsPrefixMap().containsKey( "" ) ) {
-         target.setNsPrefix( "", modelToAdd.getNsPrefixURI( "" ) );
+   /**
+    * Defensively merge a model into an existing target model. This means:
+    * <ul>
+    *    <li>Prefixes are only added when they are not already present, i.e., a model won't overwrite the empty prefix of the target model</li>
+    *    <li>Statements that have a blank node (including RDF lists) as their object won't be added</li>
```
georgschmidtdumont, line 215: Won't this cause the merged model to miss certain statements, i.e. the merged model will be incomplete?

atextor: You are correct for RDF in general, but for BAMM this should not happen. However, you had me thinking and the complete distinction is not necessary any more with the new resolution method; because a model is never merged more than once any more, so we can not end up with situations such as duplicate "?s bamm:properties ?o" statements. I have removed the corresponding section and simplified the JavaDoc.

georgschmidtdumont (approved):

core/sds-aspect-model-resolver/src/main/java/io/openmanufacturing/sds/aspectmodel/resolver/AspectModelResolver.java
chris-volk: Thanks for supporting this bug fix, Georg!

PR #218, opened 2022-09-08
Custom model resolution in bamm-cli

### LINKED ISSUE #98: [Task] Allow custom model resolvers in bamm-cli
When an Aspect Model is loaded, URNs of all used model elements need to be resolved and the referenced element needs to be defined in the same file or in some other source. This is called the model resolver. The mechanism to provide "some other source" to the model resolver is an API and the file-system based resolution is one implementation of this API. When using the SDK components in a Java application, custom resolution strategies can be provided, but for the bamm-cli, currently no such possibility exists. There should be a way to do this.

I would like to have a command line switch that takes the name/path of an external command, which can be called with the URN to be resolved as an argument and which prints the resulting Turtle to stdout. In other words, an additional resolution strategy should be implemented which spawns an additional external operating system process with the given URN as its first argument. The process's standard output should be captured and parsed using `io.openmanufacturing.sds.aspectmodel.resolver.services.TurtleLoader`. This way the extension can be flexibly done in any programming language/script language, including complex logic if necessary. Adding this resolution option to the existing bamm-cli commands can possibly be done using [picocli mixins](https://picocli.info/#_mixins).

Consider this example:
The user can write a custom resolver script, for example as a shell script:

```sh
# myresolver.sh
# Reads the model element URN as argument $1 (e.g. urn:bamm:com.example:1.0.0#MyEntity),
# converts it into a relative file system path (in this case "com/example/1.0.0/MyEntity.ttl")
# and "cat"s the corresponding file:
cat $(echo $1 | sed -e 's/urn:bamm://g;s/:/\//g;s/\#/\//g;s/$/.ttl/')
```

Then to work with an Aspect Model that references "MyEntity", for example validate it, `bamm-cli` can be called with the resolver script as an argument, allowing the model loading/resolution process to execute the resolver when it encounters a model element URN that is not present in the current model file. Example model:

```turtle
# MyAspect.ttl
@prefix : <urn:bamm:com.example:1.0.0#> .
@prefix bamm: <urn:bamm:io.openmanufacturing:meta-model:2.0.0#> .
@prefix bamm-c: <urn:bamm:io.openmanufacturing:characteristic:2.0.0#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

:MyAspect a bamm:Aspect ;
  bamm:properties ( :myProperty ) .

:myProperty a bamm:Property ;
  bamm:characteristic :MyEntityCharacteristic .

:MyEntityCharacteristic a bamm-c:SingleEntity ;
  # this is where urn:bamm:com.example:1.0.0#MyEntity is referenced, but it is not
  # defined in this file
  bamm:dataType :MyEntity .
```

Executing the bamm-cli would then look something like this:
```sh
java -jar bamm-cli-x.y.z.jar aspect MyAspect.ttl validate --custom-resolver myresolver.sh
```

Provide a .jar File with the Java implementation of the interface. However, this is difficult to maintain and test and would require regular updates to the custom resolver whenever the sds-sdk API changes.

RaMisess:
Fixes #98


tools/bamm-cli/src/test/java/io/openmanufacturing/sds/DelegatingCommandResolver.java
```diff
@@ -0,0 +1,37 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import io.openmanufacturing.sds.aspectmodel.resolver.AspectModelResolver;
+import io.openmanufacturing.sds.aspectmodel.resolver.FileSystemStrategy;
+import io.openmanufacturing.sds.aspectmodel.resolver.services.VersionedModel;
+import io.openmanufacturing.sds.aspectmodel.urn.AspectModelUrn;
+import io.vavr.control.Try;
+
+// Utility class to help test custom model resolution mechanism. It simply wraps the FileSystemStrategy, delegates the resolution to it,
```
atextor, line 26: Turn into JavaDoc

documentation/developer-guide/modules/tooling-guide/pages/bamm-cli.adoc
```diff
@@ -75,13 +81,17 @@ The available options and their meaning can also be seen in the help text of the
...
+| _--custom-resolver_ : use an external resolver for the resolution of the model elements |
+.3+| aspect <model> to schema           | Generate JSON schema for an Aspect Model                       | `bamm aspect AspectModel.ttl to schema`
                                         |  _--output, -o_ : output file path (default: stdout)           |
-.3+| aspect <model> to aas              | Generate Asset Administration Shell (AAS) submodel template for an Aspect Model  | `bamm aspect AspectModel.ttl to aas`
+| _--custom-resolver_ : use an external resolver for the resolution of the model elements |
+.4+| aspect <model> to aas              | Generate Asset Administration Shell (AAS) submodel template for an Aspect Model  | `bamm aspect AspectModel.ttl to aas`
                                         |  _--output, -o_ : output file path (default: stdout)           |
                                         |  _--format, -f_ : output file format (xml or aasx, default: xml)|
+| _--custom-resolver_ : use an external resolver for the resolution of the model elements |
 |===
```
atextor, line 96: I think a separate (short) section on the custom resolver is required here. It needs to describe the basic contract when someone wants to implement their own custom resolver.

atextor (changes requested):

atextor (approved):

PR #222, opened 2022-09-09
Successfully validate RDF models containing multiple Aspects

### LINKED ISSUE #221: [BUG] Validation of models fails for input models with > 1 Aspect
**Describe the bug**
In #199, validation was made more robust by parsing a model upon successfull validation, to prevent false positive validation results. However, when an RDF model containing more than one Aspect is passed to the validator, this will raise an `InvalidRootElementCountException`, because upon parsing, exactly one Aspect must be in the RDF model (unlike upon validation).
Validation of such models should still succeed.

**Where**
AspectModelValidator

atextor:
Fixes #221


atextor:
> Hi @atextor, the changes themselves look good to me. I do have question however. Shouldn't the validation provide information about the effect of having more than one Aspect model defined in an RDF model, given the fact that further use of this RDF model will run into issues? For example when generating code.

The validation itself must work with any amount of Aspects (or other types of model elements, that is), and this is also the previous behaviour that is being restored with this PR. The validator should be able to validate all models (or fragments thereof) such as an RDF model that contains only the definition of a BAMM Entity, or an RDF model that contains the closure of multiple Aspect Models.
The AspectModelLoader works as before and will throw the exception if more than one Aspect is defined in the model. If you argue "if it validates it should also always work for artifact generation" you are not wrong, but this must be fixed in the artifact generation APIs/CLI/AspectModelLoader and is actually a different issue.

georgschmidtdumont (approved):

PR #226, opened 2022-09-19
Adapted changes in case there's more than one constraint on a Trait

### LINKED ISSUE #223: [BUG] incorrect return return type used in generated getConstraint method in StaticConstraintProperty
**Describe the bug**
When generating a StaticMetaClass implementation for an element which contains a Property with a Trait, an implementation of the `getConstraint` method defined in the `ConstraintProperty` interface is provided. The generated implementation of the `getConstraint` method returns only the first constraint attached to the corresponding Trait.

Example:
```
@Override
public DefaultRegularExpressionConstraint getConstraint() {
	return (DefaultRegularExpressionConstraint) ((Trait) getCharacteristic().get()).getConstraints()
			.iterator().next();
}
```
The return type of the `getConstraint` method must be updated to return a list of Constraints instead of only one. The generation of the StaticMetaClass must be updated to return a list with all Constraints attached to the corresponding Trait.

Example:
```
@Override
public List<DefaultConstraint> getConstraints() {
	return ((Trait) getCharacteristic().get()).getConstraints();
}
```

**Where**
`sds-aspect-model-java-generator`

DvSmi:
Adapted changes in case there's more than one constraint on a Trait, so it will return all constraints.

fixes #223

### DIFF
--- a/core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
+++ b/core/sds-aspect-model-java-generator/src/main/resources/java-static-class-property-lib.vm
@@ -67,15 +67,15 @@
     #if( $util.hasContainerType( $property ) )
         $codeGenerationConfig.getImportTracker().importExplicit( $StaticConstraintContainerProperty )
         #set( $containedType = $util.getCharacteristicJavaType( $property, $codeGenerationConfig ) )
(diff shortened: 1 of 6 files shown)


atextor (approved):
LGTM

PR #244, opened 2022-10-28
Improve validation API

### LINKED ISSUE #201: [Task] Improve model validation API
Currently, the validator API provides methods to validate a complete loaded Aspect Model (i.e., evaluation of the BAMM SHACL shapes for a model). While this works, this lacks desirable features:
* There should be an API to perform per-element validation: Tools such as the Aspect Model Editor should be able to validate not only a complete model (because this can take multiple 100s of MS and up to multiple seconds) but also a single element. The API should analyse the which shapes apply to a given model element and evaluate only those.
* Validation messages for SHACL-built-ins should be more helpful: While shapes that specify a particular part of the meta model semantics can provide their own `sh:message`, the messages of violations of basic SHACL constraint components (e.g. `sh:nodeKind`, `sh:minCount` etc.) should be so helpful that they can be passed directly to a modeler. This means they should contain all necessary context (such as: which property on which node was checked, what were the allowed values, what are the actual values) and be as specific as possible. For example, when the shape specifies `sh:datatype rdf:langString` but the value is of type `xsd:string`, the message should be something along the lines of "The value X on property Y is missing a language tag".

The API could be as simple as `validate(Resource element)` (where the model is retrieved from `element.getModel()`).

atextor:
Fixes #201. 
This PR overhauls the Aspect Model validation API. This is backed by a customized SHACL engine implementation.

- It adds the possibility to validate either a complete resolved model or only a given element of the model using `AspectModelValiator#validateElement()`.
- It introduces a new object model for validation results. While the previous model (`ValidationError`) provided only info about focusNode, resultPath, value and severity, the new model provides the interface `Violation` and specific implementations for all possible types of SHACL violations, including SPARQL and JavaScript Constraint violations. `Violation` provides a good human-readable error message, an error code and access to the full validation context including the resource that failed validation, the SHACL shape that triggered the violation (with all of its constraints) and the originating offending RDF statements.
- It removes Topbraid SHACL as core validation dependency (it remains a Maven dependency though, see below). Current versions of Topbraid SHACL do not support the SHACL JavaScript extensions any more. For this reason, this PR provides an implementation on top of Graal JavaScript. This works both in the regular executable bamm-cli Jar and should also work in a native image when https://github.com/OpenManufacturingPlatform/sds-sdk/issues/140 is addressed.
- `ValidationReport` and is subclasses are marked deprecated for removal. Since they depend on Topbraid SHACL, this dependency is kept for now. When `ValidationReport` et al. is removed, the dependency can be removed as well. Instead of a `ValidationReport`, validation returns just a `List<Violation>` with an empty list indicating successfull validation.
- Two classes to render a `List<Violation>` to human-readable text are provided (one with the core violation message, and one which provides all validation context information). They are used in the bamm-cli `aspect validate` command: one by default instead of the old `ValidationReport`, the other by adding the `--details` switch


core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/LessThanOrEqualsViolation.java
```diff
@@ -0,0 +1,37 @@
...
+   public static final String ERROR_CODE = "ERR_LESS_THAN_OR_EQUALS";
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return String.format( "Property %s on %s must have a value that is less than or equal to that of %s: %s must less than %s",
```
RaMisess, line 29: %s must **be** less than %s

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/LessThanViolation.java
```diff
@@ -0,0 +1,37 @@
...
+   public static final String ERROR_CODE = "ERR_LESS_THAN";
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return String.format( "Property %s on %s must have a value that is less than that of %s: %s must less than %s",
```
RaMisess, line 29: %s must **be** less than %s

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/PatternViolation.java
```diff
@@ -0,0 +1,34 @@
...
+   public static final String ERROR_CODE = "ERR_PATTERN";
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return String.format( "Property %s on %s has value %s, this must match the pattern %s but it does not",
```
RaMisess, line 26: Property %s on %s has value %s, which does not match the required pattern %s.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/UniqueLanguageViolation.java
```diff
@@ -0,0 +1,36 @@
...
+   public static final String ERROR_CODE = "ERR_DUPLICATE_LANGUAGE";
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return String.format( "Property %s on %s uses language tag that has has been used already: %s",
```
RaMisess, line 28: Property %s on %s uses language tag that <s>has<s> has been used already: %s

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/NotViolation.java
```diff
@@ -0,0 +1,37 @@
...
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return context.property().isPresent() ?
+            String.format( "Expected violation of constraint %s on %s on %s, but did not occur", negatedConstraint.name(), propertyName(), elementName() ) :
```
RaMisess, line 29: Expected violation of constraint %s on %s on %s, but **it** did not occur

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/NotViolation.java
```diff
@@ -0,0 +1,37 @@
...
+            String.format( "Expected violation of constraint %s on element %s, but did not occur", negatedConstraint.name(), elementName() );
```
RaMisess, line 30: Expected violation of constraint %s on element %s, but **it** did not occur

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/MinLengthViolation.java
```diff
@@ -0,0 +1,34 @@
...
+   public static final String ERROR_CODE = "ERR_MIN_LENGTH";
+
+   @Override
+   public String errorCode() {
+      return ERROR_CODE;
+   }
+
+   @Override
+   public String message() {
+      return String.format( "Property %s on %s has length %s, but its length must be greater than or equal to %s",
```
RaMisess, line 26: General remark on all error message, not just this one: as they are nicely formulated full sentences, they should end with a dot. All our SHACL messages do. Let's keep the style consistent across all error messages.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/ModelFixer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.function.BiFunction;
+
+import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ModelFactory;
+
+import io.openmanufacturing.sds.aspectmodel.shacl.fix.Fix;
+import io.openmanufacturing.sds.aspectmodel.shacl.fix.ReplaceValue;
+
+public class ModelFixer implements BiFunction<Model, List<Fix>, Model> {
```
georgschmidtdumont, line 25: I am sure I am missing something, but as far as I can tell, this class is never used.

core/sds-aspect-model-validator/src/test/java/io/openmanufacturing/sds/aspectmodel/shacl/ShaclValidatorTest.java
```diff
@@ -0,0 +1,1737 @@
...
+      final Violation finding = violations.get( 0 );
+      assertThat( finding ).isInstanceOf( ClassTypeViolation.class );
+      final ClassTypeViolation violation = (ClassTypeViolation) finding;
+      assertThat( violation.context().element() ).isEqualTo( element );
+      assertThat( violation.context().shape().attributes().uri() ).hasValue( namespace + "MyShape" );
+      assertThat( violation.propertyName() ).isEqualTo( ":testProperty" );
+      assertThat( violation.elementName() ).isEqualTo( ":Foo" );
+      assertThat( violation.actualClass().getURI() ).isEqualTo( "http://example.com#SomethingElse" );
+      assertThat( violation.allowedClass().getURI() ).isEqualTo( "http://example.com#TestClass2" );
+      assertThat( violation.message() ).isNotEmpty();
```
georgschmidtdumont, line 159: Any reason you do not check the correctness of the message?

core/sds-test-aspect-models/src/main/resources/invalid/bamm_2_0_0/io.openmanufacturing.test/1.0.0/InvalidPreferredNameDatatype.ttl
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2021 Robert Bosch Manufacturing Solutions GmbH
```
georgschmidtdumont, line 1: Why did you not add this test model for BAMM version 1.0.0?

tools/sds-aspect-model-maven-plugin/src/main/java/io/openmanufacturing/sds/aspectmodel/AspectModelMojo.java
```diff
@@ -93,12 +95,12 @@ private void handleFailedModelResolution( final Try<VersionedModel> failedModel
 
       // Another exception, e.g. syntax error. Let the validator handle this
       final AspectModelValidator validator = new AspectModelValidator();
-      final ValidationReport report = validator.validate( failedModel );
-      throw new MojoExecutionException( report.toString(), loadModelFailureCause );
+      final List<Violation> violations = validator.validateModel( failedModel );
```
georgschmidtdumont, line 98: IMHO the option to print the detailed output for any possible violations should be available here as well.

tools/sds-aspect-model-maven-plugin/src/main/java/io/openmanufacturing/sds/aspectmodel/Validate.java
```diff
@@ -39,9 +41,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
       final Set<Try<VersionedModel>> resolvedModels = loadAndResolveModels();
       for ( final Try<VersionedModel> versionedModel : resolvedModels ) {
-         final ValidationReport report = validator.validate( versionedModel );
-         if ( !report.conforms() ) {
-            throw new MojoFailureException( report.toString() );
+         final List<Violation> violations = validator.validateModel( versionedModel );
+         if ( !violations.isEmpty() ) {
+            throw new MojoFailureException( new ViolationFormatter().apply( violations ) );
```
georgschmidtdumont, line 46: IMHO the option to print the detailed output for any possible violations should be available here as well.

core/sds-aspect-meta-model-resolver/pom.xml
```diff
@@ -106,6 +106,76 @@
...
+                                    </excludes>
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <!-- Download updated meta model shape files -->
+            <!-- This refers to the state of the shape files after merging of https://github.com/OpenManufacturingPlatform/sds-bamm-aspect-meta-model/pull/176 -->
```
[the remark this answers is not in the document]
atextor: Yes. The change is already merged on the meta model project's main branch, but a release will most likely not be done soon.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/ModelFixer.java
atextor: This is correct, this is a class that is intended to be used by consumers that want to apply a `List<Fix>` to a given model. As long as only the Fix API is place and most violations have no actual `fixes()` implementation, I'd wait with documenting this class in the developer documentation.

core/sds-aspect-model-validator/src/test/java/io/openmanufacturing/sds/aspectmodel/shacl/ShaclValidatorTest.java
atextor: The reason was not to break tests when the messages are changed, i.e., a violation would not be invalid just because it's message text changes. But since this also tests the message interpolation, it's probably good idea to do it.

core/sds-test-aspect-models/src/main/resources/invalid/bamm_2_0_0/io.openmanufacturing.test/1.0.0/InvalidPreferredNameDatatype.ttl
atextor: The current implementation of the AspectModelResolver that also used in the test suite will automatically upgrade BAMM 1.0.0-based models to 2.0.0 so the shapes won't match the model. This should probably be fixed but is out of scope for this PR IMO.

tools/sds-aspect-model-maven-plugin/src/main/java/io/openmanufacturing/sds/aspectmodel/AspectModelMojo.java
atextor: Done

tools/sds-aspect-model-maven-plugin/src/main/java/io/openmanufacturing/sds/aspectmodel/Validate.java
atextor: Done

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/violation/MinLengthViolation.java
atextor: Done

RaMisess (approved):

PR #266, opened 2022-11-22
Fix bug in the calculation of synthetic element names.

### LINKED ISSUE #225: [BUG] validation fails if model contains other node types
We are using sds-sdk version 2.0.4 and validation using io.openmanufacturing.sds.aspectmodel.validation.services.AspectModelValidator fails if the model contains other nodes with edges to Aspect Model Element nodes.
We are suspecting this might be the code issue :
Module [sds-aspect-meta-model-java](https://github.com/OpenManufacturingPlatform/sds-sdk/tree/main/core/sds-aspect-meta-model-java)  [io.openmanufacturing.sds.metamodel.loader.MetaModelBaseAttributes](https://github.com/OpenManufacturingPlatform/sds-sdk/blob/main/core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java)
```
private static String getSyntheticName( final Resource modelElement, final Model model, final BAMM bamm ) {
      Resource parentModelElement = model.listStatements( null, null, modelElement ).next().getSubject();
      while ( parentModelElement.isAnon() ) {
         parentModelElement = model.listStatements( null, null, parentModelElement ).next().getSubject();
      }      
      ....
```      
If an Aspect Model Element node has 2 edges pointing to it, sometimes the code takes the the wrong edge and in our case second call fails as there is no upper edge.

RaMisess:
Fixes #225

### DIFF
--- a/core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
+++ b/core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
@@ -26,6 +26,7 @@ import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.RDFNode;
 import org.apache.jena.rdf.model.Resource;
 import org.apache.jena.rdf.model.Statement;
+import org.apache.jena.rdf.model.StmtIterator;
 import org.apache.jena.vocabulary.RDF;
 
 import com.google.common.collect.Streams;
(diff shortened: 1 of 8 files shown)


core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
```diff
@@ -219,6 +219,31 @@ private static String getSyntheticName( final Resource modelElement, final Model
...
+   // We have to be careful when searching for the parent nodes with a regular name - the "listStatements" API returns the matching nodes
+   // in no particular order; with some very specific models this could lead to non-deterministic behavior.
+   // In the following very simplified example we are looking for ":NumberList" as the parent of "_:blankNode", but could get the anonymous node [] instead.
+   // [
+   //  aux:contains _:blankNode ;
+   // ] .
+   // :NumberList a bamm-c:List ;
+   //    bamm-c:elementCharacteristic _:blankNode .
+   // _:blankNode a bamm-c:Trait ;
+   private static Resource getNamedParent( final Resource modelElement, final Model model ) {
```
atextor, line 231: Please add unit test(s) that verify the correct behaviour.

atextor (changes requested):

core/sds-aspect-meta-model-java/src/main/java/io/openmanufacturing/sds/metamodel/loader/MetaModelBaseAttributes.java
RaMisess: Done

atextor (approved):

PR #309, opened 2023-02-13
Improve display of error messages (error line numbers and error text)

RaMisess:


core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -0,0 +1,358 @@
...
+   // some statements (like anonymous nodes or lists) can be handled out-of-order, so we remember them to not go over them twice
+   private final Set<Statement> seen = new HashSet<>();
+   private List<Statement> candidateStatements;
+
+   // The parsed model does not contain all the original tokens ( braces in lists, semicolons between statements etc.). But as we want to achieve
+   // as nice and natural formatting as possible, we look at the available information to achieve the proper spacing.
+   private List<Integer> knownPositions;
+
+   public RustLikeFormatter() {
+      textFormatter = new PlainTextFormatter();
```
atextor, line 65: I'd rather call the other constructor here, i.e. `this( new PlainTextFormatter() )`.
Also please add a comment to the constructor that explains that it will default to a PlainTextFormatter.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -0,0 +1,358 @@
...
+
+   public RustLikeFormatter( final IRdfTextFormatter textFormatter ) {
+      this.textFormatter = textFormatter;
+   }
+
+   public String constructDetailedMessage( final RDFNode highlight, final String message ) {
+      return constructDetailedMessage( highlight, message, highlight.getModel() );
+   }
+
+   public String constructDetailedMessage( final RDFNode highlight, final String message, Model rawModel ) {
```
atextor, line 76: The arguments should all be final, for necessary reassignment should be done using a new variable.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -0,0 +1,358 @@
...
+
+      textFormatter.reset();
+      statements.forEach( this::formatStatement );
+      return textFormatter.getResult();
+   }
+
+   // model.listStatements() returns the statements in random order, but we want to format them as they appear in the source document,
+   // so reconstruct the original sequence
+   private void sortSequentially( final List<Statement> statements ) {
+      statements.sort( this::documentOrder );
```
atextor, line 142: This will break if the list happens to be immutable. The method is called multiple times with results of Jena's `iterator.toList()` call, which seems to return mutable lists, but we should not rely on this. I'd change `void sortSequentially` into `List<Statement> sortSequentially` and use for example Guava's ordering: `Ordering.from(comparator).sortedCopy(list)` (documentOrder would have to be turned into a Comparator).
Other solution: Create a new ArrayList from the statements list if (and only if) it's unmodifiable (but I don't know how you would check this), then call `.sort()`.

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -0,0 +1,358 @@
...
+      }
+
+      if ( !statement.getSubject().isAnon() ) {
+         spacedIfPossible( isLastSubjectedStatement( statement ) ? "." : ";" );
+      }
+
+      return true;
+   }
+
+   // as the "whitespace" si not preserved by the parser ( in this case ';' and '.' ), we have to reconstruct it by looking at the relative positions of the
```
atextor, line 217: Typo in comment

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/validation/services/ViolationRustLikeFormatter.java
```diff
@@ -0,0 +1,212 @@
...
+import io.openmanufacturing.sds.aspectmodel.shacl.violation.Violation;
+
+public class ViolationRustLikeFormatter extends ViolationFormatter {
+
+   private final RustLikeFormatter formatter;
+
+   private final Model rawModel;
+
+   public ViolationRustLikeFormatter() {
+      rawModel = null;
```
atextor, line 55: If I understand correctly, if the ViolationRustLikeFormatter is created using this constructor, `null` will be passed to `RustLikeFormatter#constructDetailedMessage`. Is this intentional? If yes, either the model argument to the method should be made `Optional<Model>` or be annotated `javax.annotation.Nullable`.

atextor (changes requested):

core/sds-aspect-model-validator/src/main/java/io/openmanufacturing/sds/aspectmodel/validation/services/ViolationRustLikeFormatter.java
RaMisess: Yes, this is intentional. In many cases (especially unit tests) we do not have compound models (merged from several source files), so the rawModel does not need to be specified explicitly via this parameter, it can be deduced directly from the "highligh node".
Added the @Nullable annotation to the method parameter.

PR #312, opened 2023-02-17
Fix AAS mapping rule for collections

andreas-schilling:


andreas-schilling:
OK, I just noticed that the split of `SubModelElementCollection` is new to v3 of the AAS spec. Prior to that, `SubModelElementCollection` has been used for both collections (in the classic meaning of a collection) and structs. For v3 it now has been split into `SubModelElementList` for collections (again, in the classic meaning of a collection) and `SubModelElementCollection` for structs.
So, this PR is only correct once v3 is fully adopted.

atextor:
@andreas-schilling Please create an issue where you describe the problem, then reference the issue from the PR ("Fixes #xxx").

andreas-schilling:
Fixes #314

bs-jokri (approved):
LGTM

PR #317, opened 2023-03-01
the bugfix multiple instantiation abstract entities

### LINKED ISSUE #343: [BUG] Multiple instantiation of Entities when loading Aspect Models
**Describe the bug**
When an Aspect Model includes two Entities that extend the same ComplexType (it does not matter if the extended ComplexType is an `Entity` or an `AbstractEntity`), the two Entities will have references to different instances of `ComplexType`. If the `ComplexType` has child elements, they will also be instantiated twice.
Instead, for both Entities, the attribute _extends should refer to the same `ComplexType`.

**Where**
esmf-aspect-meta-model-java

ysrbo:
PR fixes duplicated instantiation of Abstract entities when loading aspect models

Fixes #343


core/esmf-aspect-meta-model-java/src/test/java/org/eclipse/esmf/metamodel/loader/AspectModelLoaderTest.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2021 Robert Bosch Manufacturing Solutions GmbH
```
atextor, line 2: 2023

core/esmf-test-aspect-models/src/main/resources/valid/samm_1_0_0/org.eclipse.esmf.test/1.0.0/AspectWithMultipleEntitiesSameExtend.ttl
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright (c) 2021 Robert Bosch Manufacturing Solutions GmbH
```
atextor, line 2: 2023

atextor (changes requested):

PR #355, opened 2023-05-08
Fix the problem with the generation of Json payloads.

### LINKED ISSUE #316: [BUG] AspectModelJsonPayloadGenerator fails sporadically with java.lang.IndexOutOfBoundsException
**Describe the bug**
Using io.openmanufacturing.sds.aspectmodel.generator.json.AspectModelJsonPayloadGenerator for generating fails from time to time fails with java.lang.IndexOutOfBoundsException. The full stacktrace is attached.

**Where**
sds-aspect-model-document-generators-2.1.3.jar class io.openmanufacturing.sds.aspectmodel.generator.json.AspectModelJsonPayloadGenerator

**Additional context**
Main "suspect" is the usage of io.openmanufacturing.sds.metamodel.impl.DefaultComplexType.instances.
The declaration of the field:
```
protected static WeakHashMap<AspectModelUrn, ComplexType> instances = new WeakHashMap<>();
```
has at least 2 problems: it is a weak hash map and it is a static field on a non-synchronized collection.

- Attached zip with tests and models:
Use test class: org.example.GenerationTest.  The bug should be reproducible if running the test with -Xmx200M. The test is single-threaded.
[org.example.jsonpayload.generation.test.zip](https://github.com/eclipse-esmf/esmf-sdk/files/10851516/org.example.jsonpayload.generation.test.zip)
- Stacktrace:
[stacktrace.txt](https://github.com/eclipse-esmf/esmf-sdk/files/10851580/stacktrace.txt)

RaMisess:
Fixes #316


documentation/developer-guide/modules/tooling-guide/examples/GenerateJsonPayload.java
```diff
@@ -14,13 +14,14 @@
...
 // tag::imports[]
+
+import java.io.File;
+import java.io.IOException;
+
 import org.eclipse.esmf.aspectmodel.generator.json.AspectModelJsonPayloadGenerator;
 import org.eclipse.esmf.aspectmodel.resolver.AspectModelResolver;
 import org.eclipse.esmf.metamodel.Aspect;
 import org.eclipse.esmf.metamodel.loader.AspectModelLoader;
-// end::imports[]
```
atextor, line 21: This is required (at the end of the imports list), because the Antora docs build uses it to select parts of the example file

atextor (changes requested):

atextor (approved):

PR #362, opened 2023-05-30
Fix evaluation of sh:node in sh:PropertyShapes

### LINKED ISSUE #361: [BUG] sh:node inside of property shapes is not correctly evaluated
**Describe the bug**
In the SHACL validator of the esmf-aspect-model-validator module, using sh:node inside a property shape is not validated correctly. The node shape that is referred to by sh:node should be evaluated against the value node(s) that are reachable via the property shape's sh:path attribute.

**Where**
esmf-aspect-model-validator, ShaclValidator and ShapeLoader

atextor:
Fixes #361

### DIFF
--- a/core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/RustLikeFormatter.java
+++ b/core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/RustLikeFormatter.java
@@ -43,12 +43,14 @@ import com.google.common.collect.Ordering;
 
 /**
  * Rust-like message formatter. Formatted messages look something like the following example:
- *
+ * <br/><br/>
+ * <pre>
  * ---> Error at line 11 column 20
  *    |
  * 11 |   :testProperty [ a :SomethingElse ]
  *    |                     ^^^^^^^^^^^^^^ Property ':testProperty' on ':Foo' has type ':SomethingElse', but only ':TestClass2' is allowed.
  *    |
+ * </pre>
  */
 public class RustLikeFormatter {
 
@@ -228,13 +230,13 @@ public class RustLikeFormatter {
       final SmartToken nodePosition = extractToken( node );
       if ( null == nodePosition ) {
          // ugly special case: Jena internally replaces the RDF keyword 'a' with 'rdf:type' without position information
-         if ( node.asNode().equals( NodeConst.nodeRDFType ) ) {
+         if ( NodeConst.nodeRDFType.equals( node.asNode() ) ) {
             spacedIfPossible( "a " );
          }
-         if ( node.asNode().equals( NodeConst.nodeTrue ) ) {
+         if ( NodeConst.nodeTrue.equals( node.asNode() ) ) {
             spacedIfPossible( "true" );
          }
-         if ( node.asNode().equals( NodeConst.nodeFalse ) ) {
+         if ( NodeConst.nodeFalse.equals( node.asNode() ) ) {
             spacedIfPossible( "false" );
          }
          return true;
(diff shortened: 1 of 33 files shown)


core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/constraint/NodeConstraint.java
```diff
@@ -14,20 +14,30 @@
...
 import org.eclipse.esmf.aspectmodel.shacl.violation.EvaluationContext;
 import org.eclipse.esmf.aspectmodel.shacl.violation.Violation;
 
 /**
  * Implements <a href="https://www.w3.org/TR/shacl/#NodeConstraintComponent">sh:node</a>
- * @param shape the node shape this sh:node refers to
+ * @param targetShape the node shape this sh:node refers to
  */
-public record NodeConstraint(Shape.Node shape) implements Constraint {
+public record NodeConstraint(Optional<Path> path, Shape.Node targetShape) implements Constraint {
```
KobOp, line 30: Should the Optional parameter be placed after the mandatory ones?

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -231,6 +231,12 @@ private boolean formatNode( final RDFNode node ) {
          if ( node.asNode().equals( NodeConst.nodeRDFType ) ) {
             spacedIfPossible( "a " );
```
KobOp, line 232: Is a space behind 'a' intended?

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/constraint/NodeConstraint.java
atextor: Nice catch, you are right

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/RustLikeFormatter.java
atextor: In this case yes, because there will always be a space after the 'a ' (= rdf:type) token when formatting, because it can only appear as the predicate in a triple and is followed by some object.

Yauhenikapl (approved):

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/RustLikeFormatter.java
```diff
@@ -231,6 +231,12 @@ private boolean formatNode( final RDFNode node ) {
...
          }
+         if ( node.asNode().equals( NodeConst.nodeTrue ) ) {
```
MelleD, line 234: I would recommend to use the constant first 
NodeConst.nodeTrue.equals(node.asNode())

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/ShaclValidator.java
```diff
@@ -128,7 +125,8 @@ public List<Violation> validateModel( final VersionedModel model ) {
    private Map<Resource, List<Shape.Node>> findSparqlTargets( final Model model ) {
       final Map<Resource, List<Shape.Node>> resourceShapes = new HashMap<>();
       for ( final Shape.Node shape : targetSparqlShapes() ) {
-         final List<Resource> shapeTargets = querySparqlTargets( model, shape.attributes().targetSparql().get() );
+         final List<Resource> shapeTargets = querySparqlTargets( model, shape.attributes().targetSparql().orElseThrow( () ->
+               new RuntimeException( "SPARQL node shape is missing a target SPARQL expression" ) ) );
```
MelleD, line 129: I would not throw a RuntimeException. Maybe a own exception or Illegal exception?

MelleD (approved):

Issue #371, opened 2023-06-21
[BUG] Font.createFont() method doesn't delete tmp file +~JF<<random_number>>.tmp after request completes.


richaashara:
**generatePng()** method inside **AspectModelDiagramGenerator class** invokes, **Font.createFont()** method. which creates a tmp font file with name pattern  +~JF<<random_number>>.tmp but doesn't delete it after the request is completed.

I am using JDK 17.
I am using this SDK in a Java application and running through a docker container. So, the docker container memory is keeping increasing sequentially after each request. As these files are going to be stored inside the container's tmp.

If I run java application without container than, those files are deleting on stopping the application. 

Code snippet where this is happemning:
```java
private void generatePng(String dotInput, OutputStream output) throws IOException {
        try {
            InputStream fontStream = this.getInputStream("diagram/RobotoCondensed-Regular.ttf");

            try {
                Font f = Font.createFont(0, fontStream);
                GraphicsEnvironment ge = GraphicsEnvironment.getLocalGraphicsEnvironment();
                ge.registerFont(f);
            } catch (Throwable var7) {
                if (fontStream != null) {
                    try {
                        fontStream.close();
                    } catch (Throwable var6) {
                        var7.addSuppressed(var6);
                    }
                }

                throw var7;
            }

            if (fontStream != null) {
                fontStream.close();
            }
        } catch (FontFormatException var8) {
        }

        MutableGraph g = (new Parser()).read(dotInput);
        Graphviz graphviz = Graphviz.fromGraph(g);
        graphviz.render(guru.nidi.graphviz.engine.Format.PNG).toOutputStream(output);
    }
```

Font method decompiled implementation is as below . :
```java
private static Font[] createFont0(int fontFormat, InputStream fontStream, boolean allFonts, CreatedFontTracker tracker) throws FontFormatException, IOException {
        if (fontFormat != 0 && fontFormat != 1) {
            throw new IllegalArgumentException("font format not recognized");
        } else {
            boolean copiedFontData = false;

            try {
                final File tFile = (File)AccessController.doPrivileged(new PrivilegedExceptionAction<File>() {
                    public File run() throws IOException {
                        return Files.createTempFile("+~JF", ".tmp").toFile();
                    }
                });
                if (tracker != null) {
                    tracker.add(tFile);
                }

                int totalSize = 0;

                try {
[...]

PR #377, opened 2023-06-29
- tmp font file created from font stream.

richaashara:
This resolves the issue: https://github.com/eclipse-esmf/esmf-sdk/issues/371

### DIFF
--- a/core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
+++ b/core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
@@ -24,6 +24,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.Base64;
 import java.util.EnumMap;
 import java.util.List;
@@ -173,7 +174,7 @@ public class AspectModelDiagramGenerator {
       // To make the font available during PNG generation, it needs to be registered
       // in Java Runtime's graphics environment
       try{
-         File tmpFontFile = gnererateTmpFontFile();
+         final File tmpFontFile = generateTmpFontFile();
          final Font f = Font.createFont( Font.TRUETYPE_FONT, tmpFontFile );
          final GraphicsEnvironment ge = GraphicsEnvironment.getLocalGraphicsEnvironment();
          ge.registerFont( f );
@@ -185,14 +186,13 @@ public class AspectModelDiagramGenerator {
       final Graphviz graphviz = Graphviz.fromGraph( g );
       graphviz.render( guru.nidi.graphviz.engine.Format.PNG ).toOutputStream( output );
    }
-   private File gnererateTmpFontFile() throws IOException {
-      File tempFontFile =  new File(TMP_FONT_FILE);
-      if (!tempFontFile.exists()){
-         try (InputStream fontStream =  getInputStream(FONT_FILE)){
-            try (OutputStream output = new FileOutputStream(tempFontFile, false)){
-               fontStream.transferTo(output);
+   private File generateTmpFontFile() throws IOException {
+      File tempFontFile = new File( System.getProperty( "java.io.tmpdir" ) + "/aspect-model-diagram.tmp" );
+      if ( !tempFontFile.exists() ){
+         try ( final InputStream fontStream = getInputStream( FONT_FILE );
+              final OutputStream output = new FileOutputStream( tempFontFile, false ) ){
+               fontStream.transferTo( output );
             }
-         }
       }
       tempFontFile.deleteOnExit();
       return  tempFontFile;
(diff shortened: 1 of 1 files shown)


core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
```diff
@@ -180,7 +185,18 @@ private void generatePng( final String dotInput, final OutputStream output ) thr
       final Graphviz graphviz = Graphviz.fromGraph( g );
       graphviz.render( guru.nidi.graphviz.engine.Format.PNG ).toOutputStream( output );
    }
-
+   private File gnererateTmpFontFile() throws IOException {
```
atextor, line 188: There is a typo in this method name; should be `generateTmpFontFile`.

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
atextor, line 1: Please adapt the code style/formatting to the rest of the code:
* every variable should be marked `final` where possible
* there should be a space after `(` and before `)` for method calls and constructor calls, `try`-with-resources etc. (except when they are empty)

atextor (changes requested):

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
```diff
@@ -180,7 +185,18 @@ private void generatePng( final String dotInput, final OutputStream output ) thr
...
+      File tempFontFile =  new File(TMP_FONT_FILE);
```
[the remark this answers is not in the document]
richaashara: 1. I will make the tempFontFile variable final and will use, `System.getProperty("java.io.tmpdir")` to get the tmp path so that it can work on all OS.
2. Creating a file using `Files.createTempFile( "aspect-model-diagram", null ).toFile()` will create a file with a random name. so, rather than creating the file using the temp file create function can we create a proper file but inside tmp? so, for each request, we don't need a new file and can use an existing one?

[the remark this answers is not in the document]
richaashara: 1. I will make the tempFontFile variable final and will use, `System.getProperty("java.io.tmpdir")` to get the tmp path so that it can work on all OS.
2. Creating a file using `Files.createTempFile( "aspect-model-diagram", null ).toFile()` will create a file with a random name. so, rather than creating the file using the temp file create function can we create a proper file but inside tmp? so, for each request, we don't need a new file and can use an existing one? 
   `final File tempFontFile = new File(System.getProperty("java.io.tmpdir")+"/aspect-model-diagram/Font-RobotoCondensed-Regular.tmp" );`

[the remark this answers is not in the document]
atextor: Since the content of this file will always be the same (it's always the font) this sounds reasonable. But you would have to use `File.separator` or construct it using `java.nio.file.Path` to prevent using the platform-dependent `/`.

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
```diff
@@ -80,6 +83,8 @@ public String getArtifactFilename( final String aspectName, final Locale languag
    static final String GET_ELEMENT_NAME_FUNC = "urn:samm:org.eclipse.esmf.samm:function:2.0.0#getElementName";
    static final String GET_NAMESPACE_FUNC = "urn:samm:org.eclipse.esmf.samm:function:2.0.0#getNamespace";
 
+   private static final  String TMP_FONT_FILE = "/tmp/Font-RobotoCondensed-Regular.tmp";
```
atextor, line 86: This is not used anymore, please remove

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/diagram/AspectModelDiagramGenerator.java
```diff
@@ -180,7 +186,17 @@ private void generatePng( final String dotInput, final OutputStream output ) thr
       final Graphviz graphviz = Graphviz.fromGraph( g );
       graphviz.render( guru.nidi.graphviz.engine.Format.PNG ).toOutputStream( output );
    }
-
+   private File generateTmpFontFile() throws IOException {
+      File tempFontFile = new File( System.getProperty( "java.io.tmpdir" ) + "/aspect-model-diagram.tmp" );
```
atextor, line 190: Please use `File.separator` instead of the hardcoded `/` (to have it work on Windows)

atextor (changes requested):

atextor (approved):
Looks good to me!

chris-volk:
Thanks for your contribution @richaashara !

richaashara:
Hello,
May I know, when this fix will be released. So, I can update the esmf-sdk dependency version in my project.

PR #448, opened 2023-10-17
Bugfix/375 javabeans bool spec fix

### LINKED ISSUE #375: [BUG] isGetters of generated Java POJOs get skipped by Jackson when Optional booleans are used
**Describe the bug**
If an Aspect model contains optional booleans, the Java POJO generator builds an accessor like for regular booleans using the isGetter naming scheme:
```java
public Optional<Boolean> isHidden() {
  return this.hidden;
}
```

During visibility detection Jackson however does two checks:
* check the name matches the isGetter convention (which it does)
* check whether the return type is `boolean` (not the case)

This results in the field becoming omitted from the payload.

**Where**
`java-pojo-getter-lib.vm`

**Possible fixes**
Multiple solutions exist to solve this issue:

1. use the default getter naming scheme in such a case (in this example name the method `getHidden`)
2. add an additional `@JsonProperty` annotation (simply with the same name applied) if the `boolean` is wrapped inside an `Optional`

Most likely the first option is not what most people would expect. The second option might be confusing to see when someone looks at the generated code, because at first glance it does not make sense to apply a name "override" that does not change anything in terms of property naming.

**Immediate fix**
As of now, the problem is most easily fixed using a Jackson mixin with a custom Jackson Module:

```java
public abstract class MyPojoMixin {
   @JsonProperty(value = "hidden")
   abstract public Optional<Boolean> isHidden();
}

...

setMixInAnnotation( MyPojo.class, MyPojoMixin.class );
```

ysrbo:
Fixed getter name generator according to JavaBeans specification

Fixes #375

### DIFF
--- a/core/esmf-aspect-model-java-generator/pom.xml
+++ b/core/esmf-aspect-model-java-generator/pom.xml
@@ -57,6 +57,10 @@
          <artifactId>record-builder-processor</artifactId>
          <scope>provided</scope>
       </dependency>
+      <dependency>
+         <groupId>org.apache.jena</groupId>
+         <artifactId>jena-core</artifactId>
+      </dependency>
 
       <!-- Test dependencies -->
       <dependency>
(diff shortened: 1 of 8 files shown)


Yauhenikapl (approved):

Yauhenikapl:
Don't forget update branch

Thanks,
Best Regards!

atextor (approved):

PR #459, opened 2023-10-19
Fix #432 several aas issues hindering import to aas package explorer

### LINKED ISSUE #432: [BUG] CLI SAMM2AASX errors when generating AAS Submodel Template
**Describe the bug**
I had a look at https://github.com/eclipse-tractusx/sldt-semantic-models/tree/main/io.catenax.batch/2.0.0/gen and wanted to edit it with the aasx package explorer. This is not yet supported.
Note: I do not know which version of the CLI generator is used in Catena-X at this point in time.

I had a look at the generated xml file and found the following issues. For importing into aasx package explorer **fixing 2) and 6)** is the most important, then **4).** Then it can be imported.

1) Remove "Optional and brackets from value" (see also 2)
```
              <aas:key>
                <aas:type>GlobalReference</aas:type>
                <aas:value>Optional[urn:samm:io.catenax.batch:2.0.0#Batch]</aas:value>
              </aas:key>
```
or
`<aas:id>Optional[urn:samm:io.catenax.batch:2.0.0#Batch]/submodel</aas:id>`
etc.

2) Remove embeddedDataSpecification for AAS. There no data specifications defined for AAS. A SAMM model is not a data specification.

Note: This is sufficient to load it to the aasx package explorer in the first place!

3) recommendation to not add optional elements if they are not included like
`  </aas:displayName>`
`</aas:value>`

4) The data specification for concept descriptions has the value (see Part 3a V3.0)
`https://admin-shell.io/DataSpecificationTemplates/DataSpecificationIec61360/3/0`
i.e. instead of
```xml
<aas:embeddedDataSpecification>
          <aas:dataSpecification>
            <aas:type>ExternalReference</aas:type>
            <aas:keys>
              <aas:key>
                <aas:type>GlobalReference</aas:type>
                <aas:value>Optional[urn:samm:io.catenax.batch:2.0.0#Batch]</aas:value>
              </aas:key>
            </aas:keys>
          </aas:dataSpecification>
          <aas:dataSpecificationContent>
            <aas:dataSpecificationIec61360>
```

we have

```xml
<aas:embeddedDataSpecification>
          <aas:dataSpecification>
            <aas:type>ExternalReference</aas:type>
            <aas:keys>
              <aas:key>
                <aas:type>GlobalReference</aas:type>
                <aas:value>https://admin-shell.io/DataSpecificationTemplates/DataSpecificationIec61360/3/0</aas:value>
              </aas:key>
            </aas:keys>
          </aas:dataSpecification>
          <aas:dataSpecificationContent>
            <aas:dataSpecificationIec61360>
```

5)  use "en" for language code consistently, i.e. change
```xml
              <aas:shortName>
                <aas:langStringShortNameTypeIec61360>
                  <aas:language>EN</aas:language>
                  <aas:text>Batch</aas:text>
                </aas:langStringShortNameTypeIec61360>
```

6) please add a submodel reference to the AAS for the SMT. Otherwiese submodel will not correctly be loaded to aasx package explorer

```xml
	  			<aas:submodels>
				<aas:reference>
					<aas:type>ModelReference</aas:type>
					<aas:keys>
						<aas:key>
							<aas:type>Submodel</aas:type>
							<aas:value>urn:samm:io.catenax.batch:2.0.0#Batch</aas:value>
						</aas:key>
					</aas:keys>
				</aas:reference>
			</aas:submodels>
```

6) It seems "https://admin-shell.io/aas/3/0 AAS.xsd" a \ is missing before AAS.xsd?

`<aas:environment xmlns:aas="https://admin-shell.io/aas/3/0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://admin-shell.io/aas/3/0 AAS.xsd">`

7)  There is a prefix "id_" for each idShort. Remove prefix "id_" from generated idShort names

8) There is a value "Unknown" for many Properties.
  
image
This is not a valid xs:dateTime value. In these cases please just do not add any value. Instead the example value "2022-02-04T14:48:54" should be added usw. or no value at all

**Where**
CLI

Edit @atextor: XML formatting

### LINKED ISSUE #462: [BUG] AAS Generator extracts scalar property values with property default toString() implementation
**Describe the bug**
Scalar values are currently only transferred with the default implementation of property toString() method causing some metadata being added as well.
```
<aas:valueReferencePair>
  <aas:value>DefaultScalarValue[value=raw material, typeUri='DefaultScalar[metaModelVersion=SAMM_2_1_0,
    urn='http://www.w3.org/2001/XMLSchema#string']']</aas:value>
  <aas:valueId>
  <aas:type>ModelReference</aas:type>

This is definitely wrong; this is just the toString() representation of the ScalarValue (which includes metadata such as its type and the used meta model version). Since it's an xsd:string, in this case it should look like <aas:value>raw material</aas:value>. I think for most types you can just use the lexical representation (i.e., use toString() on the value), but I'm not sure about xsd:hexBinary/xsd:base64Binary and especially about rdf:langString).
```

**Where**
core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAASVisitor.java

**Screenshots**

**Additional context**

bs-jokri:
Fixes #432. Fixes #462.


bs-jokri:
# Current state of work

## Still open issues
3. recommendation to not add optional elements if they are not included like
8. This is not a valid xs:dateTime value. In these cases please just do not add any value. Instead the example value "2022-02-04T14:48:54" should be added usw. or no value at all


## Completed
2. Remove embeddedDataSpecification for AAS. There no data specifications defined for AAS. A SAMM model is not a data specification.
4. The data specification for concept descriptions has the value (see Part 3a V3.0)
7. There is a prefix "id_" for each idShort. Remove prefix "id_" from generated idShort names
5. use "en" for language code consistently, i.e. change
6. please add a submodel reference to the AAS for the SMT. Otherwiese submodel will not correctly be loaded to aasx package explorer
Current shape:
```
 <aas:submodels>
        <aas:reference>
          <aas:type>ModelReference</aas:type>
          <aas:keys>
            <aas:key>
              <aas:type>Submodel</aas:type>
              <aas:value>urn:samm:io.catenax.batch:2.0.0#Batch/submodel</aas:value>
            </aas:key>
          </aas:keys>
        </aas:reference>
      </aas:submodels>
```
1. Remove "Optional and brackets from value" (see also 2)
6. It seems "https://admin-shell.io/aas/3/0 AAS.xsd" a \ is missing before AAS.xsd?
https://github.com/eclipse-aas4j/aas4j/issues/186



## Possible further problems identified while implementing
Question, are they real problems?

```
 <aas:valueReferencePair>
                    <aas:value>DefaultScalarValue[value=raw material, typeUri='DefaultScalar[metaModelVersion=SAMM_2_1_0, urn='http://www.w3.org/2001/XMLSchema#string']']</aas:value>
                    <aas:valueId>
                      <aas:type>ModelReference</aas:type>
```

atextor:
> Question, are they real problems?
>
> ```xml
> <aas:valueReferencePair>
>   <aas:value>DefaultScalarValue[value=raw material, typeUri='DefaultScalar[metaModelVersion=SAMM_2_1_0,
>     urn='http://www.w3.org/2001/XMLSchema#string']']</aas:value>
>   <aas:valueId>
>   <aas:type>ModelReference</aas:type>
> ```

This is definitely wrong; this is just the `toString()` representation of the `ScalarValue` (which includes metadata such as its type and the used meta model version). Since it's an xsd:string, in this case it should look like `<aas:value>raw material</aas:value>`. I think for most types you can just use the lexical representation (i.e., use `toString()` on the `value`), but I'm not sure about xsd:hexBinary/xsd:base64Binary and especially about rdf:langString).

bs-jokri:
Quickfix added for 
It seems "https://admin-shell.io/aas/3/0 AAS.xsd" a \ is missing before AAS.xsd?
https://github.com/eclipse-aas4j/aas4j/issues/186

However, this only works for xml serialization AASX files still have the old schemaLocation structure. Could be fixed by implementing unzipping of the aasx archive, applying the implemented fix to the content.xml file and zip again.

bs-jokri:
> > Question, are they real problems?
> > ```
> > <aas:valueReferencePair>
> >   <aas:value>DefaultScalarValue[value=raw material, typeUri='DefaultScalar[metaModelVersion=SAMM_2_1_0,
> >     urn='http://www.w3.org/2001/XMLSchema#string']']</aas:value>
> >   <aas:valueId>
> >   <aas:type>ModelReference</aas:type>
> > ```
> 
> This is definitely wrong; this is just the `toString()` representation of the `ScalarValue` (which includes metadata such as its type and the used meta model version). Since it's an xsd:string, in this case it should look like `<aas:value>raw material</aas:value>`. I think for most types you can just use the lexical representation (i.e., use `toString()` on the `value`), but I'm not sure about xsd:hexBinary/xsd:base64Binary and especially about rdf:langString).

Created ticket for that
https://github.com/eclipse-esmf/esmf-sdk/issues/462

bs-jokri:
Still open issue: 
recommendation to not add optional elements if they are not included like

Problem here is that selectively filling or omiting elements in the resulting AAS document is hard to achieve with the API of the AAS4j library as it provides a builder-based API, which does not allow to selectively add thing. So fixing this would require to create a stub-object first, which is subsequently filled with the missing data. 

Also. Still open question: What are optional elements? How to determine, for which the change needs to be applied.

bs-jokri:
@atextor @andreas-schilling The PR fixes all but one issue mentioned in #432, which I think is a bigger implementation effort. I would therefore suggest to create a new issue for the "optional elements" issue and merge the fixes so that the generated XML files can be used within the AASPackageExplorer

bs-jokri:
BTW I tested the fixes at the AAS generator with the batch model of Catena-X, which resulted in a successful import in the AASPackageExplorer
aaspackageexplorter

bs-jokri:
@andreas-schilling checking the screenshot above with all the unknown entries, I have the impression that the DefaultMapper not using the provided exampleValues of a property at all. Could that be the case?

Yauhenikapl (approved):
Looks good!

andreas-schilling:
> @andreas-schilling checking the screenshot above with all the unknown entries, I have the impression that the DefaultMapper not using the provided exampleValues of a property at all. Could that be the case?

Yes, that very much looks like. I think the `getPropertyValue` method here (https://github.com/eclipse-esmf/esmf-sdk/blob/main/core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/Context.java#L178) should look something like this: (untested straight from my brain)

```java
public String getPropertyValue( final String defaultValue ) {
   return getRawPropertyValue().map( valueNode -> valueNode.asText( defaultValue ) )
         .or( () -> property.getExampleValue().map( ScalarValue::getValue ).map( Object::toString ) )
         .orElse( defaultValue );
}
```

The question however is, whether the mapping process should differentiate between the variant using a payload and the one which doesn't on a global level or not. What I mean with this is: with the above fix the behaviour would be that payload values are used when a payload is given and example values if this is either not the case or no value was found. This could lead to the situation that you get a mix of actual payload values and example values, which I could imagine is not desirable. Especially for optional values this would be highly intransparent.
This should be easily fixable though by changing the order of the chain of optionals above: first try to retrieve an example value (but only, when no payload is given, else return `Optional.empty()` right away), then try to retrieve an actual payload value, then fall back to default.

bs-jokri:
> > @andreas-schilling checking the screenshot above with all the unknown entries, I have the impression that the DefaultMapper not using the provided exampleValues of a property at all. Could that be the case?
> 
> Yes, that very much looks like. I think the `getPropertyValue` method here (https://github.com/eclipse-esmf/esmf-sdk/blob/main/core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/Context.java#L178) should look something like this: (untested straight from my brain)
> 
> ```java
> public String getPropertyValue( final String defaultValue ) {
>    return getRawPropertyValue().map( valueNode -> valueNode.asText( defaultValue ) )
>          .or( () -> property.getExampleValue().map( ScalarValue::getValue ).map( Object::toString ) )
>          .orElse( defaultValue );
> }
> ```
> 
> The question however is, whether the mapping process should differentiate between the variant using a payload and the one which doesn't on a global level or not. What I mean with this is: with the above fix the behaviour would be that payload values are used when a payload is given and example values if this is either not the case or no value was found. This could lead to the situation that you get a mix of actual payload values and example values, which I could imagine is not desirable. Especially for optional values this would be highly intransparent. This should be easily fixable though by changing the order of the chain of optionals above: first try to retrieve an example value (but only, when no payload is given, else return `Optional.empty()` right away), then try to retrieve an actual payload value, then fall back to default.

Seems like we need an architecture decision and probably also a new ticket to handle the discussion of that issue.

atextor:
I have merged main into the branch and fixed conflicts. I've also added the fix for #462, IMO this belongs here. Regarding the remaining open point ("recommendation to not add optional elements if they are not included like"); I can not reproduce a problem here. The builder methods for e.g. displayName are called with empty lists in case the source Aspect Model does not contain preferredNames, and they correspondingly also don't appear in the resulting aasx/aas.xml. Therefore I consider this PR ready.

atextor (approved):

PR #465, opened 2023-10-24
Fix display of example values in HTML

### LINKED ISSUE #323: [BUG] HTML Doc: enumerations displayed as "value.getValue"
**Describe the bug**
In case of enumeration with string values html documentation does not display the value but the operation $value.getValue() per value in the enumeration. The example value is "$property.getExampleValue().get().getValue()"

Instead the values should be displayed themselves.

**Where**
Indicate the location of the bug (e.g., which component of the SDK, which API call, which command etc.).

**Screenshots**
grafik

grafik

atextor:
Fixes #323


Yauhenikapl (approved):
Looks so good!

PR #504, opened 2024-01-16
Update OpenAPI generation documentation in regards to tenantId

### LINKED ISSUE #503: [BUG] OpenAPI generation documentation is out of date
* Generate an OpenAPI description from an Aspect Model, with and without setting `--resource-path`

Without `--resource-path` set, the default resource path is `/{tenant-id}/name-of-the-aspect`.
With `--resource-path` set, the resource path is exactly what is given by the option.
The documentation however claims, that `{tenant-id}` is always part of the resource path, which is wrong.

java-aspect-tooling.adoc

atextor:
Fixes #503


Yauhenikapl (approved):
Looks good!

PR #516, opened 2024-01-26
Extend java static meta model accessors

### LINKED ISSUE #515: [Task] Extend Java static meta model with accessor functionality
The Java static meta model is an intuitive way to work with aspects, where either explicit or generic usage of the properties is required. However, with the respective property instances at hand, a developer still can't use it to execute generic _actions_ on them that would require to access the underlying data of an actual instance of the Aspect DTO.
A simple example would be the following:

* We have an application with multiple Aspects, each of them contains multiple properties
* The application wants to offer a new API, which provides clients the option to perform a data export that does only include a reduced set of properties (which the client also can control to some extend)
* For the sake of simplicity we assume that the data export simply prints to console, but the concept is the same

The basic algorithm now would use some collection of properties to decide, which data to export:

```java
List<StaticProperty<?>> toExport = List.of(MetaAspect.PROPERTY_ONE, MetaAspect.PROPERTY_TWO);

toExport.stream().map(property -> // ???).forEach(propertyValue -> System.out.println(propertyValue));
```

The spot marked with `???` is the missing functionality: except for reflection, there is no other option to use the property to access the actual value of an instance of the Aspect DTO.

The generated static meta model classes need to have another method (e.g. `getValue()`, that takes an isntance of the enclosing class (Aspect, Entity, ...) as a parameter and simply executes the respective getter on it. During code generation all context (class, getter name, ...) is known.
With such a solution in place, the above code could be written like this:

```java
toExport.stream().map(property -> property.getValue(aspectInstance)).forEach(propertyValue -> System.out.println(propertyValue));
```

It would be possible to provide data accessor functionality through an external class that uses the meta model and the respective instances, but that would rather complicate the usage of the static meta model and also be way less intuitive.

andreas-schilling:
Fixes #515


andreas-schilling:
While I consider the PR to be fine as it is, there is also an alternative way to implement this.
The accessor method could be extracted to its own interface
```java
public interface PropertyAccessor<C, T> {
   T getValue( C object );
}
```
so that the static property changes to
```java
public abstract class StaticProperty<C, T> extends DefaultProperty implements PropertyAccessor<C, T>
```

This doesn't change much in a first stage, however if we think a bit further into the direction of chained properties, computed properties or other ad-hoc defined metamodel elements (either at compile- or at runtime), this would allow writing static-like code also on non-static metamodel elements, that might have been dynamically loaded during runtime.

This could be handy when operating on certain properties in a truly generic application scenario where the meta information of some part is known to the developer (like _"I want to operate on a property that represents a temperature measurement, independent from the Aspect/Entity surrounding it and I also don't even know what meta model elements I get hold of while the application runs"_).
I know that this is a tad abstract and also only gains the same amount of type safety when it is implemented using runtime checks that guarantee, that the implemented bits can only be applied to data that actually matches the ad-hoc meta information.
Still, the change as of now would be rather small and I could adjust the PR accordingly.

WDYT?

atextor (approved):

atextor:
> While I consider the PR to be fine as it is, there is also an alternative way to implement this. The accessor method could be extracted to its own interface
> 
> ```java
> public interface PropertyAccessor<C, T> {
>    T getValue( C object );
> }
> ```
> 
> so that the static property changes to
> 
> ```java
> public abstract class StaticProperty<C, T> extends DefaultProperty implements PropertyAccessor<C, T>
> ```
> 
> This doesn't change much in a first stage, however if we think a bit further into the direction of chained properties, computed properties or other ad-hoc defined metamodel elements (either at compile- or at runtime), this would allow writing static-like code also on non-static metamodel elements, that might have been dynamically loaded during runtime.
> 
> This could be handy when operating on certain properties in a truly generic application scenario where the meta information of some part is known to the developer (like _"I want to operate on a property that represents a temperature measurement, independent from the Aspect/Entity surrounding it and I also don't even know what meta model elements I get hold of while the application runs"_). I know that this is a tad abstract and also only gains the same amount of type safety when it is implemented using runtime checks that guarantee, that the implemented bits can only be applied to data that actually matches the ad-hoc meta information. Still, the change as of now would be rather small and I could adjust the PR accordingly.
> 
> WDYT?

I like the idea of having this and I'd be fine with both approaches, adjusting this PR or create a new one with this change. I approved this PR as-is and I leave it up to you if you'd like to change it again or rather create a separate one. The only thing I'd ask for in addition would be some user documentation (in java-aspect-tooling.adoc) that explains how to use this feature.

Issue #340, opened 2023-04-02
[Task] Add Model information to the Openapi Spec


matgnt:
This is a feature request with low priority.

**Is your task related to a problem? Please describe.**
When I generate the code from the openapi spec, I want some basic model meta information included, e.g. the org name, model name, version, maybe more...

**Describe the solution you'd like**
Probably an enum with that kind of information, like 'XXModel.org' or 'XXModel.version' or 'XXModel.fullName'

**Describe alternatives you've considered**
I'm currently defining those additionally even though, the information would be available from the source of what is used to create the openapi spec.

Thanks for the tools,
Matthias Binzer

matgnt:
Yes, probably a good idea to add the information to the metadata info object.
Since I'm not sure if all code generators would use this information and 'how', maybe to following could be an option, too.

Example TTL file:
https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.assembly_part_relationship/1.1.1/AssemblyPartRelationship.ttl#L33
(gen folder contains the generated files...)

A potential option could be something like:
```
components:
  schemas:
    ModelPrefixName:
      type: string
      enum:
      - "urn:bamm:io.catenax.assembly_part_relationship:1.1.1#"
    ModelName:
      type: string
      enum:
      - "AssemblyPartRelationship"
    ModelFullName:
      type: string
      enum:
      - "urn:bamm:io.catenax.assembly_part_relationship:1.1.1#AssemblyPartRelationship"

```
I think there is no other proper way to add some 'const' information to an openapi spec.

This would generate the enums that could be used. Disadvantage would be, when a version changes, also the enum names would change.

matgnt:
Yes, you're right, it is not used in any endpoint here. It is an endpoint generated from another Openapi Spec. E.g. the one from the AAS Registry.
I use the string "urn:bamm:io.catenax.assembly_part_relationship:1.1.1#AssemblyPartRelationship"
when creating an endpoint of an AAS in the registry. There, you need to provide the `semanticId` and therefore, currently it is a little bit of manual work :-)

matbmoser:
In my understanding would be good to have a json which can contain a generic node structure including "name", "description", and other attributes. Also would be important to search for this attributes because of this a "ref" for the path to this node in tree.

This would be my proposal:

![image](https://github.com/eclipse-esmf/esmf-sdk/assets/71728767/08e9b853-eef1-4ff6-8bb1-3082b3983f74)

[...]

PR #539, opened 2024-03-11
Add Model information to the Openapi Spec

Yauhenikapl:
Fixes: #340

### DIFF
--- a/core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
+++ b/core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
@@ -231,6 +231,7 @@ public class AspectModelJsonSchemaVisitor implements AspectVisitor<JsonNode, Obj
 
    public JsonNode visitAspectForOpenApi( final Aspect aspect ) {
       addDescription( rootNode, aspect, locale );
+      addXSammAspectModelUrn( rootNode, aspect );
       return visitHasProperties( aspect, rootNode );
    }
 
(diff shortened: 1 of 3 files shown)


atextor (changes requested):
It looks good so far. It would be great to have an `x-samm-aspect-modle-urn` for the schemas as well (containing the URN of the Entity).

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
```diff
@@ -671,4 +672,8 @@ private ObjectNode addDescription( final ObjectNode node, final NamedElement des
 
       return node;
    }
+
+   private void addXSammAspectModelUrn( final ObjectNode node, final NamedElement describedElement ) {
+      node.put( "x-samm-aspect-model-urn", describedElement.getAspectModelUrn().map( Object::toString ).orElse( "" ) );
```
atextor, line 677: I think it could happen that the NamedElement's AspectModelUrn is empty, in that case no `x-samm-aspect-model-urn` should be added IMO. Maybe like this:

```suggestion
      describedElement.getAspectModelUrn().ifPresent( urn -> node.put( "x-samm-aspect-model-urn", urn.toString() );
```

atextor (changes requested):

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/jsonschema/AspectModelJsonSchemaVisitor.java
Yauhenikapl: Hm, got it.
I expect, if we don't have urn, we don't need to make this field in the spec. 

Okay,
I am going to fix this,

Thanks,
Best Regards!

atextor (approved):

PR #568, opened 2024-04-16
Fix checkstyle action token configuration

atextor:
The config was copied from the [action's
documentation](https://github.com/marketplace/actions/checkstyle-for-java), but
this is wrong: As can be seen in other places of the action, the secrets
variable must be upper case.


Yauhenikapl (approved):
LGTM!

ysrbo (approved):

PR #573, opened 2024-05-10
Create PUT/PATCH/POST methods for OpenApi

Yauhenikapl:
Changes:
- update OpenAPI generator;
- add new flag for OpenAPI generation to commands;
- update tests.


ysrbo (approved):

atextor (changes requested):
I think the following changes are necessary in this PR:
* PUT/POST/PATCH should be selectable separately. Sometimes it might make sense to have only a combination or subset of them.
* There should be corresponding test(s) for SammCli also, to check if the command line switch works as expected
* The test in the Maven Plugin should also have useful assertions, currently only the test case pom (generate-openapi-spec-json-pom-valid-aspect-model.xml) is adjusted, but not the test method(s)

atextor (approved):
LGTM!

PR #583, opened 2024-05-28
Add semanticId to all list elements

### LINKED ISSUE #522: [Task] CLI aasx Generation - add semanticId to all elements
Up to now the CLI for aasx and aas.xml generation only adds semanticId for the elements for which there is a Concept Description generated and for the submodel itself. However, the aspect model also provides semanticIds for SubmodelElementList (SML) and SubmodelCollection (SMC): these should be added as well. It is an external reference in this case.

`				
                    <type>ExternalReference</type>

				<keys>
					<key>
						<type>GlobalReference</type>
						<value>urn:samm:io.....</value>
					</key>

				</keys>
`

Yauhenikapl:
## Description

Add semanticId to Collection elements.

Fixes #522

## Type of change
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
+++ b/core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
@@ -387,11 +387,11 @@ public class AspectModelAasVisitor implements AspectVisitor<Environment, Context
 
    private Reference buildReferenceForCollection( final String submodelId ) {
       final Key key = new DefaultKey.Builder()
-            .type( KeyTypes.SUBMODEL_ELEMENT_COLLECTION )
+            .type( KeyTypes.GLOBAL_REFERENCE )
             .value( submodelId )
             .build();
       return new DefaultReference.Builder()
-            .type( ReferenceTypes.MODEL_REFERENCE )
+            .type( ReferenceTypes.EXTERNAL_REFERENCE )
             .keys( key )
             .build();
    }
@@ -557,7 +557,7 @@ public class AspectModelAasVisitor implements AspectVisitor<Environment, Context
             new DefaultSubmodelElementList.Builder()
                   .idShort( property.getName() )
                   .semanticId( buildReferenceForCollection( collection.getAspectModelUrn().isPresent()
-                        ? collection.getAspectModelUrn().get().getUrn().toString() : collection.getName() ) )
+                        ? collection.getAspectModelUrn().get().getUrn().toString() : property.getName() + "Characteristic" ) )
                   .typeValueListElement( AasSubmodelElements.DATA_ELEMENT )
                   .displayName( LangStringMapper.NAME.map( property.getPreferredNames() ) )
                   .description( LangStringMapper.TEXT.map( property.getDescriptions() ) )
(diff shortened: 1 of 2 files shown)


core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
```diff
@@ -385,6 +385,17 @@ private Reference buildReferenceForSubmodel( final String submodelId ) {
             .build();
    }
 
+   private Reference buildReferenceForCollection( final String submodelId ) {
+      final Key key = new DefaultKey.Builder()
+            .type( KeyTypes.SUBMODEL_ELEMENT_COLLECTION )
+            .value( submodelId )
+            .build();
+      return new DefaultReference.Builder()
+            .type( ReferenceTypes.MODEL_REFERENCE )
```
atextor, line 394: I think this needs to be "external reference"

core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
```diff
@@ -385,6 +385,17 @@ private Reference buildReferenceForSubmodel( final String submodelId ) {
             .build();
    }
 
+   private Reference buildReferenceForCollection( final String submodelId ) {
+      final Key key = new DefaultKey.Builder()
+            .type( KeyTypes.SUBMODEL_ELEMENT_COLLECTION )
```
atextor, line 390: I think this needs to be "global reference"

core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
```diff
@@ -545,6 +556,8 @@ private <T extends Collection> Environment visitCollectionProperty( final T coll
       final SubmodelElementBuilder builder = property ->
             new DefaultSubmodelElementList.Builder()
                   .idShort( property.getName() )
+                  .semanticId( buildReferenceForCollection( collection.getAspectModelUrn().isPresent()
```
atextor, line 559: If the collection has no URN it is only locally valid for the Property that refers to it. In that case we can't use the collection's local name a semantic ID; instead we need to either (1) add no semantic ID at all for this element (probably the cleanest solution) or (2) create an identifier for the semantic ID that is based on the Property's URN, e.g. by concatenating `propertyName + "Characteristic"`.

atextor (changes requested):

core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
Yauhenikapl: Yes, thanks for the clarification. I think, 2nd option is more acceptable, because in the task means, that some semantic Id has to be provided.

Thanks!

BirgitBoss: Hi: I also prefer option (1) as @atextor proposed. We should not add IDs that are not existing. 
When writing the task I  forgot to consider the anoymous characteristics. SemanticId is not mandatory so it is fine to not add a semanticId in these cases. (However, it would be best practice to not use anonymous characteristic for collections if aasx should be generated out of it)

atextor (approved):

BirgitBoss:
Unfortunately, there seems to be no samm-cli-2.8.0-windows-x86_64.zip offered as part of the release?

PR #588, opened 2024-06-07
579 implement mapping elements namespaces files

### LINKED ISSUE #579: [Task] Implement mapping of model namespaces to files
In https://github.com/eclipse-esmf/esmf-semantic-aspect-meta-model/issues/181 the relationships between model elements, namespaces and files are properly defined. In this issue, the Java API should be adapated to reflect these changes. In particular, the following changes need to be implemented:

* The esmf-aspect-meta-model-java needs to introduce the proper abstractions for
    * `AspectModel` as an abstraction that contains a number of namespaces and can therefore be considered the "context" or "top level" element for a loaded set of elements,
    * `ModelElementGroup` which indicates something is a collection of ModelElements (e.g., an `AspectModel`, a `ModelFile` and a `ModelNamespace`,
    * `ModelFile` which logically represents a single file; it has a 1:1 releationship to a Jena `Model`
* AspectModelLoader and AspectModelResolver need to adjusted to allow loading instances of the new abstractions.
* The problematic `VersionedModel` class which contains only a merged RDF model after model resolution (without any provenance information or distinction of the orignal sources) needs to be removed and its usage replace with corresponding new abstractions.
* By extension, the `AspectContext` class becomes redundant and needs to be removed.
* APIs taking either `VersionedModel` or `AspectContext` as an input need to be adjusted to take an object of a more fitting class instead, such as `ModelElement`, `Aspect`, `List<Aspect>` etc.
* The concept of supporting multiple SAMM versions simultaneously on RDF level is deprecated (the original problem is adressed by automatic meta model version migration instead) and its use should be removed (i.e., every unnecessary reference of `KnownVersion` or usage of `KnownVersion` as arguments to API functionality outside the meta model version migrator).

* The new abstractions should clearly represent the relationships between files, model elements and namespaces. The classes must allow navigation between this logical model element graph.
* When changes are made to the model graph, the abstractions must allow writing back files to the file system (or their respective other source, if it is writable).
* Aspect loading and resolution should be straightforward and not involve more steps than necessary.
* If possible, interfaces for generated Java code (POJOs and static meta classes) should not change (their internal code may change of course).

ysrbo:
## Description

Implement mapping of model namespaces to files

Fixes #579

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)


atextor:
Will be fixed in separate pull request.

Issue #547, opened 2024-03-15
[Task] Extend generic Aspect to AAS conversion to allow more fine-grained customizations


andreas-schilling:
**Is your task related to a problem? Please describe.**
The current implementation of the Aspect to AAS converter is a good starting point for simple and quick results starting from an Aspect model (and potentially also Aspect data in form of a JSON payload).
If the main task is to create an AAS representation of a custom Aspect model, this is sufficient as it is. However, for many tasks it is required to be close to an IDTA specified submodel, which has strict specifications about elements like the ShortIds of elements or their Semantic Ids.
This can partially be handled by overriding the Property Mapping mechanism, but it would still not allow doing this for other elements, e.g. `SubmodelElementCollection` or `SubmodelElementList`.

**Describe the solution you'd like**
The conversion setup needs to allow more points where developers can hook in to change the behaviour.
A non-exhaustive list would be:

- Short-Id generation
- Semantic-Id generation
- Creation of each element that can appear in an AAS (e.g. `SubmodelElementCollection`)

Addditionally, the ESMF-SDK should provide default implementations as a starting point for developers, so that certain standard use cases are already covered (e.g. having a Semantic-Id generator that uses the `see` element of SAMM properties) and ready to be used.
Ideally, full customization is still only necessary for very special variants of a conversion process.

PR #598, opened 2024-06-25
Allow property mapping customizations for collection properties

andreas-schilling:
This PR implements the smaller part of #547 and now allows to customize the mapping of collection valued properties from the outside through a `PropertyMapper`.
Additionally, this PR fixes two related issues that have been found during development:

- Set the content type of BLOBs which will be used if a collection property of scalars can't be converted to a proper type. Without this, tools like the AASX explorer refuse to load the resulting AASX file
- Extend the `LangStringPropertyMapper` to also work with collection properties directly, where the single language entries are given as separate objects within a collection


core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
```diff
@@ -568,28 +570,29 @@ private <T extends Collection> Environment visitCollectionProperty( final T coll
...
-         createSubmodelElement( builder, context );
-         return context.getEnvironment();
-      } ).orElseGet( () -> {
-         createSubmodelElement( builder, context );
-         return context.getEnvironment();
-      } );
+      final SubmodelElementBuilder listBuilder =
+            tryFindPropertyMapper( context.getProperty() )
+                  .flatMap( mapper -> collection.getDataType()
+                        .map( type -> (SubmodelElementBuilder) ( Property property ) -> mapper.mapToAasProperty( type, property,
```
atextor, line 576: For nested expression code like this, please wrap after `->` to increase readability

atextor (changes requested):

core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/AspectModelAasVisitor.java
andreas-schilling: done.

atextor (approved):
LGTM!

PR #662, opened 2024-10-18
Extract esmf-native-support module from samm-cli

### LINKED ISSUE #661: [Task] Refactor native-image configuration to make it reusable
samm-cli is built using GraalVM native-image. The build requires lots of special setup, such as generation of configuration files during the build, substitution classes and build-time post-processing. In order to make esmf-sdk reusable in other projects that use GraalVM native image compilation, this should be as convenient as possible without having to recreate such setup. This is particularly useful for the Aspect Model Editor backend.

Introduce a new module that contains the native-image configuration files that are created using integration tests and necessary build-time setup. Other modules (i.e., samm-cli) and projects (i.e., AME backend) should be able to declare a dependency on this module, and be able to run their own native-image compilation with esmf-sdk core functionality "just working".

atextor:
-------

## Description

Introduces a new module, `esmf-native-support`, that can be used as a dependency
in projects that use esmf-sdk and use GraalVM native-image compilation.

Fixes #661.

## Type of change

## Checklist:


core/esmf-native-support/src/main/java/org/eclipse/esmf/nativefeatures/Native.java
```diff
@@ -0,0 +1,446 @@
...
+   public static Class<?> clazz( final String className ) {
+      try {
+         return Class.forName( className );
+      } catch ( final ClassNotFoundException exception ) {
+         throw new RuntimeException( "Could not access for reflection registry", exception );
+      }
+   }
+
+   protected static boolean isLinux() {
+      return new IsLinux().getAsBoolean();
```
Yauhenikapl, line 253: I am not sure, do we need IsMacOS?

Yauhenikapl (approved):
LGTM!

core/esmf-native-support/src/main/java/org/eclipse/esmf/nativefeatures/Native.java
atextor: We can add it if we need it in the native setup on MacOS. Until now, we've only encountered platform-specific behavior in the JDKs on Windows and Linux that we had to work around.

PR #667, opened 2024-10-24
Enable loading model with references from GitHub

### LINKED ISSUE #474: [Task] Load model with references from github
As a user of SAMM CLI / ESMF SDK I want to load an Aspect Model from a github url that may include references to other models which are also in github and validate/read it (step 1) as well as make changes to it (step 2)
so that I can work with it quickly without having to worry about manually downloading all references / cloning an entire github repository.

(Additionally, as an ideation, As a modeler I want to add multiple sources to my modeling environment config
so that I can reuse these sources for loading/resolution automatically/more conveniently.)

Example: 
- With the SAMM CLI I want to give https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.generic.digital_product_passport/3.0.0/DigitalProductPassport.ttl as an input and validate it
- The SAMM CLI shall fetch the 5 referenced namespaces starting @prefix ext-address: <urn:samm:io.catenax.shared.address_characteristic:3.0.0#> . from the corresponding locations, e.g. https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.shared.address_characteristic/3.0.0/AddressAspect.ttl (including transitive references . In the case of cyclic references - which can happen-, no endless recursion must happen)
- The validation result shall happen just as if it had been locally in the correct folder structure
- Instead of validating I may want to download the model and its references to a specified target root + correct folder structure OR as a zip. After that, validation must also be possible on the downloaded copy.
- ~~instead of downloading a single aspect model, it shall also be possible to download an entire namespaces by giving a github folder for a namespace, e.g. https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.shared.address_characteristic/3.0.0~~ (moved to #666)
- ~~As an important option for entire namespaces I want to only download the newest version of each namespace (and if older versions are referenced these must be included so that resolution of included aspect models is possible)~~ (moved to #666)

In addition to being the basis of making this accessible via Aspect Model Editor (followup story) this issue also serves as a simple demo of the SAMM CLI.

The guideline is that the with the two inputs SAMM CLI + link for Aspect Model on github, the question "is the model valid?" must be answered by an average internet user very fast and easily, minimum number of steps and questions/research (development and documentation to follow the https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

In scope of this issue:

public github repositories with referred namespaces in the correct relative folder structure (meaning: in the same repository)
SDK and SAMM CLI
~~private github repositories (also correct relative structure), use a library for authentification that works for common private github repositories~~ (note: moved to #665)

Out of scope:

- git not hosted by github.
- loading into Aspect Model Editor (this will be a linked issue)
- uploading (valid) models to github

### LINKED ISSUE #664: [BUG] Regression: Loading models with multiple references to another file fails
Load an Aspect Model that has multiple implicit references to another file, i.e., by different paths that lead to inclusion of the other file multiple times. One example of such an Aspect Model is CatenaX [Battery Pass 6.0.0](https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.battery.battery_pass/6.0.0/BatteryPass.ttl).

Actual result:
The merged RDF model in the loaded Aspect Model has duplicate assertions for all statements where the object is a blank node (e.g. `samm:properties), which leads to validation errors and processing errors.
Expected result:
Model resolution works properly even if multile include paths lead to including the same file multiple times

atextor:
## Description

With the changes in this PR, samm-cli commands no longer only take a local file
name as input for the Aspect Model, but also allow GitHub URLs and Aspect Model
URNs. Resolution from a given GitHub repository can be configured as a
resolution strategy, so it is now also possible to e.g. have an Aspect Model as
a local file that refers to elements hosted on GitHub (via usage of their
namespace; the resolution strategy will then figure out the location of the
corresponding file and retrieve it).

Fixes #474, fixes #664

## Type of change

## Checklist:


Yauhenikapl (approved):
LGTM!

PR #668, opened 2024-10-25
Add Prefix and Postfix to Java Generation

### LINKED ISSUE #538: [Task] Allow customizing the class name when generating POJOs from aspect models
Currently the name that is used when generating POJOs from aspect models is fixed to the element name.
It would be great to have more control over the name, especially to allow custom prefixes.
The background for this is the following: it is quite common for projects that use generated POJOs to have the generated layer as some kind of base and then use derived classes where appropriate, that are the actual POJOs used to implemented against:

```java
BaseMovement.java
 |_ Movement.java
```
This setup allows extended or otherwise customized functionality on POJOs to be provided through the derived classes (in this example within `Movement.java`).

To be clear, this setup of course is already possible to be implemented with the current state, however a somewhat weird naming scheme would be required to do so, as the parent classes would carry the idiomatic name and the derived classes would need some pre- or postfix.

A minimum solution would allow to add a prefix on demand, an ideal solution would allow to fully plug in a function that generates the final name to be used for the class (with the default behaving as it is right now).
Also, the solution should (at least in the most fine-grained configurable variant) allow to perform this customizing on a per-model base, so that it is possible to selectively apply a customized name. This would allow the above described implementation structure to be applied to just a bunch of models, while keeping the default for all the others, because no customized implementation layer is required for these.

Yauhenikapl:
## Description

Allow customizing the class name when generating POJOs from aspect models

Fixes #538 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaUtil.java
+++ b/core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaUtil.java
@@ -29,7 +29,6 @@ import java.util.stream.Stream;
 import org.eclipse.esmf.aspectmodel.java.exception.CodeGenerationException;
 import org.eclipse.esmf.aspectmodel.visitor.AspectStreamTraversalVisitor;
 import org.eclipse.esmf.metamodel.AbstractEntity;
-import org.eclipse.esmf.metamodel.Aspect;
 import org.eclipse.esmf.metamodel.Characteristic;
 import org.eclipse.esmf.metamodel.ComplexType;
 import org.eclipse.esmf.metamodel.Entity;
@@ -354,7 +353,7 @@ public class AspectModelJavaUtil {
       }
       return TO_CONSTANT.convert( StringUtils.capitalize( upperOrLowerCamel ) );
    }
-   
+
    public static boolean isAllUppercaseWithUnderscore( final String upperOrLowerCamel ) {
       return upperOrLowerCamel != null && upperOrLowerCamel.matches( "[A-Z0-9_]+" );
    }
@@ -561,8 +560,8 @@ public class AspectModelJavaUtil {
    }
 
    public static String generateClassName( final StructureElement element, final JavaCodeGenerationConfig config ) {
-      if ( ( !config.aspectPrefix().isBlank() || !config.aspectPostfix().isBlank() ) && element instanceof Aspect ) {
-         return config.aspectPrefix() + element.getName() + config.aspectPostfix();
+      if ( ( !config.namePrefix().isBlank() || !config.namePostfix().isBlank() ) && element.is( StructureElement.class ) ) {
+         return config.namePrefix() + element.getName() + config.namePostfix();
       }
       return element.getName();
    }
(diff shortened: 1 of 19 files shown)


core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaUtil.java
```diff
@@ -559,6 +560,13 @@ public static String genericClassSignature( final StructureElement element ) {
       return generics.isEmpty() ? "" : "<" + generics + ">";
    }
 
+   public static String generateClassName( final StructureElement element, final JavaCodeGenerationConfig config ) {
+      if ( ( !config.aspectPrefix().isBlank() || !config.aspectPostfix().isBlank() ) && element instanceof Aspect ) {
```
atextor, line 564: I think this should work not only for Aspects, but also for other structure elements (i.e., Entities and AbstractEntities):
```suggestion
      if ( ( !config.aspectPrefix().isBlank() || !config.aspectPostfix().isBlank() ) && element.is( StructureElement.class ) ) {
```

core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/JavaCodeGenerationConfig.java
```diff
@@ -29,7 +29,10 @@ public record JavaCodeGenerationConfig(
       String packageName,
       ImportTracker importTracker,
       boolean executeLibraryMacros,
-      File templateLibFile
+      File templateLibFile,
+      String aspectPrefix,
+      String aspectPostfix
```
atextor, line 34: Can we rename this to e.g. "namePrefix" and "namePostfix" (since it also applies to Entities and AbstractEntities)?

core/esmf-aspect-model-java-generator/src/test/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaGeneratorTest.java
```diff
@@ -1109,4 +1122,18 @@ public void testGenerateAspectWithMultipleInheritance() throws IOException {
...
+            .put( "testProperty", String.class )
+            .build();
+
+      final TestAspect aspect = TestAspect.ASPECT_WITH_PROPERTY;
+      final GenerationResult result = TestContext.generateAspectCode()
+            .apply( getGenerators( aspect, "Base", "Postfix" ) );
+      result.assertNumberOfFiles( 1 );
+      result.assertFields( "BaseAspectWithPropertyPostfix", expectedFieldsForAspectClass, new HashMap<>() );
+      assertConstructor( result, "BaseAspectWithPropertyPostfix", expectedFieldsForAspectClass );
+   }
```
atextor, line 1138: Please also add one more test for TestAspect.ASPECT_WITH_ENTITY and check that the prefix/postfix is also applied to the class that is generated for the Entity

atextor (changes requested):

atextor:
Please also add documentation in samm-cli.adoc and maven-plugin.adoc.

atextor (approved):

PR #670, opened 2024-11-04
Add org.graalvm.js:js dependency is missing in the fat jar

### LINKED ISSUE #663: [BUG] `samm validate` not working with samm-cli 2.9.6 on MacOS
`samm aspect SomeAspectModel.ttl validate` with the native binary

Error message: `java.lang.NullPointerException: Cannot invoke "javax.script.ScriptEngine.getBindings(int)" because "this.engine" is null`

**Input (files/screenshots)**

Apparently the `org.graalvm.js:js` dependency is missing in the fat jar that is built using warp-packer. 

Also check why the integration tests (seemingly) don't run for the MacOS native binary.

Edit: This problem also appears on other platforms when using the Java API or the executable .jar file.

Yauhenikapl:
## Description

Add org.graalvm.js:js dependency is missing in the fat jar

Fixes #663 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


atextor (approved):

core/esmf-aspect-model-validator/pom.xml
```diff
@@ -43,7 +43,8 @@
       <dependency>
          <groupId>org.graalvm.js</groupId>
          <artifactId>js</artifactId>
-         <type>pom</type>
+         <type>jar</type>
+         <version>${graalvm-version}</version>
```
MelleD, line 47: Why do you need the version and type here?

Yauhenikapl: If you install type pom, plugin will exclude this dependency from the fat jar. But in documentation this is not so clear. Regarding version, to be more consistent with current version of graalvm

MelleD: But you already declared that in parent POM. IMHO should also work without version and type

Yauhenikapl: Yes, but I tried this without type and version and maven started excluding this package from fat jar...
But, let's try some new build

MelleD: I can't understand it, it also works locally for me and also with the maven plugin.
I think the problem lies somewhere else. There is no type pom for this deps. Thats the reason why you don't need type and version because you declared it in parent pom

PR #678, opened 2024-11-17
Meta Class doesn't include exampleValue

### LINKED ISSUE #623: [BUG] Generated meta class does not include exampleValue in Properties
* Generate static Java meta class for Aspect Model

* Expected: `samm:exampleValue` for Properties are initialized in the static code with the value from the Aspect Model
* Actual: exampleValue is initialized with `Optional.empty()`

Tried with [PartAsPlanned](https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.part_as_planned/2.0.0/PartAsPlanned.ttl); the Property `catenaXId` has an example value in the model but not in the class.

Yauhenikapl:
## Description

Update Meta Class generation to include exampleValue.

Fixes #623  

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
## Checklist:

### DIFF
--- a/core/esmf-aspect-model-java-generator/src/test/java/org/eclipse/esmf/aspectmodel/java/StaticMetaModelJavaGeneratorTest.java
+++ b/core/esmf-aspect-model-java-generator/src/test/java/org/eclipse/esmf/aspectmodel/java/StaticMetaModelJavaGeneratorTest.java
@@ -24,8 +24,11 @@ import java.nio.file.Path;
 import java.time.LocalDate;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Pattern;
+
 import javax.xml.datatype.DatatypeFactory;
 import javax.xml.datatype.XMLGregorianCalendar;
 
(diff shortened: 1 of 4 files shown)


core/esmf-aspect-model-java-generator/src/test/java/org/eclipse/esmf/aspectmodel/java/StaticMetaModelJavaGeneratorTest.java
```diff
@@ -498,6 +501,34 @@ void testCharacteristicInstantiationForQuantifiableWithoutUnit() throws IOExcept
...
+      result.assertMetaModelBaseAttributesForProperties( "MetaAspectWithCollection", expectedBaseAttributes );
+
+      final String expectedTestPropertyCharacteristicConstructorCall =
+            "new DefaultCollection(MetaModelBaseAttributes.builder().withUrn(AspectModelUrn.fromUrn(NAMESPACE + \"TestCollection\"))"
+                  + ".withPreferredName(Locale.forLanguageTag(\"en\"), \"Test Collection\").withDescription(Locale.forLanguageTag(\"en\")"
+                  + ", \"This is a test collection.\").withSee(\"http://example.com/\").build(), Optional.of(new DefaultScalar"
+                  + "(\"http://www.w3.org/2001/XMLSchema#string\")), Optional.empty())";
+
+      result.assertConstructorArgumentForProperties( "MetaAspectWithCollection",
+            ImmutableMap.<String, String> builder().put( "TEST_PROPERTY", expectedTestPropertyCharacteristicConstructorCall ).build(), 1 );
```
atextor, line 529: I'm missing the assertion that looks for the presence of `Optional.of(new DefaultScalarValue("Example Value", new DefaultScalar("http://www.w3.org/2001/XMLSchema#string")))`, i.e., the thing we actually want to assert here.
`assertConstructorArgumentsForProperties` can not be used as-is for this, because it only checks the constructor of the Characteristic and it's not straightforward because the example value declaration is nested deep in the field declaration.

You could do something like this:

Add those methods in `StaticClassGenerationResult`:
```java
   private Stream<Node> getAllChildNodes( final FieldDeclaration field ) {
      return field.getChildNodes().stream().flatMap( this::nodeAndItsChildren );
   }

   private Stream<Node> nodeAndItsChildren( final Node node ) {
      return Stream.concat( Stream.of( node ), node.getChildNodes().stream().flatMap( this::nodeAndItsChildren ) );
   }
```

and then in `assertConstructorArgumentForProperties`, add an (optional) argument for the field declaration to check (which you use when calling the method from the test, passing the example value declaration), and use `getAllChildNodes(field).map(Node::toString)` if any of the nodes contains the declaration.

atextor (changes requested):

atextor (approved):

PR #712, opened 2025-02-25
Enable import and export of Namepace Packages

### LINKED ISSUE #666: [Task] Functionality to import and export namespace packages
[ADR-0009](https://github.com/eclipse-esmf/esmf-semantic-aspect-meta-model/blob/main/documentation/decisions/0009-namespace-packages.md) specifies the layout of so-called _namespaces packages_, zip files that contain an Aspect Model namespace directory structure and Aspect Model .ttl files.
A samm-cli command is required to create and read these files. This includes at least the following actions:
* Create a namespace package from an input (URN or file) and its transitive dependencies, or from a given set of namespaces and/or aspect model files.
* Import ("extract") a namespace package into a models root directory in the local file system in way that makes sense, with configurable boundary conditions (e.g., "overwrite all", "import only newer versions", "show differences",...)
* A shortcut that transparently combines the two points above in order to sync the state of one model source to another. "Transparently" here means that a user does not see or needs to interact with the intermediary .zip file. The main use case here is "import the latest state from GitHub repository X to local workspace directory Y".

For all functionality there needs to be a Java API (so that this feature can be used from the Aspect Model Editor backend) and a corresponding samm-cli command.

atextor:
## Description

Adds `package import` and `package export` commands in samm-cli

Fixes #666

## Type of change

## Checklist:


Yauhenikapl (approved):
LGTM!

PR #754, opened 2025-05-08
Fix Java code generation for StructuredValues

### LINKED ISSUE #750: [BUG] Generating code for a property, which has a Trait characteristic, results in DUPLICATE
**Observed on CLI Version: samm-cli-2.10.2-windows-x86_64**

When we generate a code (Java) for a model, which has a property(s) with a Trait-characteristic, 

the resultant code has duplicates for the properties of the Trait model.

This happens with the ESMF-CLI tool.

Command we use is:
`samm aspect .\aspects\com.bosch.datalayer.dt2cs.shared.vehiclemanagement\1.0.0\VehicleType.ttl to java -pn javaFiles`

**Input (files/screenshots)**
The model with a Trait characteristic:
Image

Duplicated Fields code:
Image

Duplicated Getters code:
Image

**NOTE:** When we remove the Trait and use only the "YearAndOptionalMonth" characteristic, the above duplicates are gone in the generated code. 

**Suggestion:** Maybe, since a Trait is used to introduce a sort of validation/verification on particular property (characteristic) **instead of** creating a new field / getter etc, as presumably the trait is handled as "second" characteristic, **to use the Trait** to set a sort of RegExp / Min-Max limits etc., to the related Characteristic code.

As per above examples, Screenshots: **instead of** having duplicate entries for the **year** and **month** fields, to include a Java Notation or other mechanism to introduce the specified by the Trait RegularExpression to validate both of the mentioned fields.

### LINKED ISSUE #753: [BUG] Generating code for a property, which has a regex constraint, results in importing two different classes with the same name
Observed on CLI Version: samm-cli-2.10.2-windows-x86_64

When we generate a code (Java) for a model, which has a property(s) with a regex constraint.

Actual: The resultant code is importing two different classes with the same name.
Expected: The resultant code should import one class and use the other’s fully qualified class name in code.

This happens with the ESMF-CLI tool.
Command we use is:
`samm aspect .\aspects\com.bosch.datalayer.dt2cs.shared.vehiclemanagement\1.0.0\VehicleType.ttl to java -pn javaFiles`

**Input (files/screenshots)**
The model with a Trait characteristic having regex constraint.
Image

imports with two different classes with the same name.
Image

Method 'compile' not found for pattern since it is taking by default
`jakarta.validation.constraints.Pattern.compile` instead of
`java.util.regex.Pattern.compile`
Image

The issue is resolved when only 1 class is used in import and for other, we use fully qualified class name.

Image

atextor:
When more than one Property with StructuredValue that refers to the same
deconstructed Properties is in scope, generated fields need to be prefixed.

Fixes #750, fixes #753


Yauhenikapl (approved):
LGTM!

Issue #798, opened 2025-08-07
[BUG] HTML generation: Nondeterministic output when Aspect names clash


atextor:
**How: List of performed steps**
* Load an Aspect Model that, including its dependencies, contains at least two `samm:Aspect`s with the same local name. 
* Run HTML generation for loaded model.

**What: Actual result vs expected result**
* Expected: Either the Aspect to use is determined from the given file name or path and the documentation for the corresponding Aspect is generated, or an error is thrown because it's not clear which Aspect to generate for.
* Actual: Nondeterministic behavior: Sometimes documentation is generated for one of the Aspects, sometimes for the other

**Where: (e.g. component, version, url, your system info)**
esmf-sdk 2.11.1

PR #814, opened 2025-09-12
SAMM2AASX Generator: wrong generated semanticId for Submodel

### LINKED ISSUE #794: [BUG] SAMM2AASX Generator: wrong generated semanticId for Submodel

A)
```
@prefix bp: <urn:samm:io.BatteryPass.Circularity:1.2.0#> .
@prefix : <urn:samm:io.admin-shell.idta.batterypass.circularity:1.2.0#> .

:Circularity a samm:Aspect ;
```

B)
```
@prefix : <urn:samm:io.admin-shell.idta.batterypass.circularity:1.2.0#> .
@prefix bp: <urn:samm:io.BatteryPass.Circularity:1.2.0#> .

:Circularity a samm:Aspect ;
```

Expected Result: in both cases the generated semanticId for the Submodel "Circularity" should be
`
urn:samm:io.admin-shell.idta.batterypass.circularity:1.2.0#Circulariy`

The order of the `@prefix` is not relevant for the resulting semanticIds.

However, in case B) the generated semanticId is

`urn:samm:io.BatteryPass.Circularity:1.2.0#Circularity`

In general it does not seem to be the last `@prefix` but there seems to be another selection rule behind... perhaps the namespace of the first property that is declared in the samm:Aspect?

Used CLI:
```
samm-cli - Semantic Aspect Meta Model Command Line Tool
Version: 2.10.3
Build date: 2025-05-08T14:47:57Z
Git commit: 192116d12a1a4c9df4e03758a6b56dfa80132cf3
```

e-filchenko-bosh:
## Description

Bug Description:
This bug reproduces inconsistently.

Reason:
When we upload AspectModels with all dependencies, we also save fileLocation as part of the context. Currently, these locations are saved without ordering. As a result, during the next step of resolving files, we extract the name from the list of Aspect names that we saved based on the ordering in fileLocations.

In this scenario, the name .../target/Aspect.ttl might be lower in the order than .../second/Aspect.ttl. Consequently, we end up using the semantic ID from the wrong URN.

This problem also affects the logic in bug #798.

Fixes #794

## Type of change

## Checklist:

- [-] I have commented my code, particularly in hard-to-understand areas
- [-] I have made corresponding changes to the documentation
- [-] My changes generate no new warnings

### DIFF
--- a/core/esmf-aspect-meta-model-java/src/main/java/org/eclipse/esmf/aspectmodel/loader/AspectModelLoader.java
+++ b/core/esmf-aspect-meta-model-java/src/main/java/org/eclipse/esmf/aspectmodel/loader/AspectModelLoader.java
@@ -589,14 +589,18 @@ public class AspectModelLoader implements ModelSource, ResolutionStrategySupport
    @Override
    public boolean containsDefinition( final AspectModelFile aspectModelFile, final AspectModelUrn urn ) {
       final Model model = aspectModelFile.sourceModel();
+      boolean result = model.contains( model.createResource( urn.toString() ), RDF.type, (RDFNode) null );
+      if ( result ) {
+         LOG.debug( "Checking if model contains {}: {}", urn, result );
+         return result;
+      }
       if ( model.getNsPrefixMap().values().stream().anyMatch( prefixUri -> prefixUri.startsWith( "urn:bamm:" ) ) ) {
-         final boolean result = model.contains(
+         result = model.contains(
                model.createResource( urn.toString().replace( AspectModelUrn.PROTOCOL_AND_NAMESPACE_PREFIX, "urn:bamm:" ) ), RDF.type,
                (RDFNode) null );
          LOG.debug( "Checking if model contains {}: {}", urn, result );
          return result;
       }
-      final boolean result = model.contains( model.createResource( urn.toString() ), RDF.type, (RDFNode) null );
       LOG.debug( "Checking if model contains {}: {}", urn, result );
       return result;
    }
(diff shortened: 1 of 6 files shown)


core/esmf-test-aspect-models/src/main/resources/valid/org.eclipse.esmf.test.ordering.dependency/1.0.0/Aspect.ttl
```diff
@@ -0,0 +1,30 @@
...
+@prefix : <urn:samm:org.eclipse.esmf.test.ordering.dependency:1.0.0#> .
+
+:Aspect a samm:Aspect ;
+   samm:description "Test dependency aspect "@en ;
+   samm:properties ( :test ) ;
+   samm:operations ( ) ;
+   samm:events ( ) .
+
+:test a samm:Property ;
+   samm:name "test property" ;
```
atextor, line 25: Please remove this line, "samm:name" was removed in SAMM 2.0.0.

atextor (changes requested):
While all changes and checks in the PR are valid and usful, I think it does not fix the reported bug. At least one problem remains samm-cli: In samm-cli's [AbstractInputHandler.java:154](https://github.com/eclipse-esmf/esmf-sdk/blob/main/tools/samm-cli/src/main/java/org/eclipse/esmf/AbstractInputHandler.java#L154), the list of loaded Aspects is filtered for the "designated name" (i.e., the name of the originally loaded input file, minus ".ttl") and the first Aspect with that name is used. But this is non-deterministic if there happen to be multiple Aspects with that name in the resolved model. In order to fix this, another test is needed in SammCliTest (using \@RepeatedTest is a good idea). A possible fix would be to have `AbstractInputHandler#expectedAspectName()` not return the local name of the Aspect as a `String`, but the full `AspectModelUrn`.

atextor (approved):

PR #824, opened 2025-11-03
Patch: Documentation for generated openAPI-files

jSchuetz88:
Corrected the formatting of tenantId to tenant-id in the documentation for the generation of openAPI-files.

## Description

I'm not sure and maybe just got the documentation wrong. The documentation uses camelCase for the attribute `tenantId` while the actual generated openAPI-specification generates `tenant-id` and seems to use kebab-case. However, as the tenant-id is only a placeholder, this patch only serves the consistency and might not be necessary.

Fixes #(number of issue in GitHub)

## Type of change

## Checklist:

Not applicable.

## Additional notes:


atextor:
Hi @jSchuetz88,
thanks for your contribution! I think a reader can understand it either way, but your change makes it more consistent, so it's welcome.

PR #856, opened 2025-12-23
Remove usage of GraalVM

### LINKED ISSUE #852: [Task] Remove usage of GraalVM
The samm-cli module makes use of GraalVM and its native-image compilation to create a native binary. Although this achieves the original goal of providing a runnable command that does not require users to install a Java runtime first, it comes with several problems:

* Compilation takes a long time; a full esmf-sdk build takes about ~1h, half of which is due to native-image compilation.
* The native-image setup introduces additional classes of failures. In addition to complation failures and regular runtime failures, there's a new class of runtime failures that _only_ occur in the native binary but not the executable .jar (e.g., when resource configuration was wrong). This requires a complex setup to catch. Our solution currently is to separately run all integration tests both with the executable .jar and the native binary.
* Creation and maintenance of the native-image configuration (reflection/JNI/resources/etc.) is complex and expensive.
    * Just to make aas4j and its rather involved reflection setup work, there's a specific [build-time only](https://github.com/eclipse-esmf/esmf-sdk/blob/main/core/esmf-native-support/src/main/java/org/eclipse/esmf/buildtime/Aas4jClassSetup.java) code that reproduces it, creates a .properties file from it, which is included in the build as a resource, to be loaded at native initialization time again.
    * Other non-trivial setup specific to third-party dependencies [is necessary](https://github.com/eclipse-esmf/esmf-sdk/tree/main/core/esmf-native-support/src/main/java/org/eclipse/esmf/nativefeatures).
    * AWT font configuration (transitive requirement for diagram generation) is broken on Windows as it requires - hardcoded in JRE - a font config file which exists in the JRE but not in the native-image. The [hack](https://github.com/eclipse-esmf/esmf-sdk/blob/main/core/esmf-native-support/src/main/java/org/eclipse/esmf/nativefeatures/DiagramFeature.java#L403-L407) to make it work is to copy the resource from the JRE into the project resources at build time.
    * AWT with native-image on MacOS [remains broken](https://github.com/oracle/graal/issues/4124). Since the build does not work, we use a different means for building the samm-cli binary on that platform (see also "solution" section below).
* There is no (straightforward) way for extending samm-cli other than at compile time. While using dynamic class loading mechanisms would be feasible in plain Java, we don't have this possibility with the native image.

We still want to provide a self-contained version of samm-cli (i.e., which does not require to first install a Java runtime). This can be achieved with projects which wrap an executable fat jar with a runtime; we already use this approach for samm-cli on MacOS using the warp-packer project.

The changes should therefore remove all GraalVM-native-image-specific parts and update the build to use warp-packer instead.

* Remove `esmf-native-support` module completely
* In root project:
    * Remove `native` profile from pom.xml
    * Remove native-related properties and `graalvm-version` from pom.xml
* In samm-cli project:
    * Remove `native` profile from pom.xml
    * Remove `packaging-type` property from pom.xml
    * Remove passing `packaging-type` 
    * Remove SammCliFeature class
    * Simplify tests
        * SammCliIntegrationTest can be removed
        * `@EnabledIfSystemProperty( named = "packaging-type", matches = "native" )` in `SammCliTest` can be removed
        * References to `agentlib:native-image-agent` and `nativeConfigPath` can be removed
* Update GitHub actions
    * Remove `graalvm/setup-graalvm` action, use `setup-java` instead
    * Update mvn calls to not run `-Pnative`
    * Bundle using warp-packer instead (see MacOS build), use [kirbylink/warp-packer](https://github.com/kirbylink/warp/releases/tag/v1.1.1) (maintained fork)
* Update documentation regarding removal of esmf-native-support and GraalVM support

e-filchenko-bosh:
## Description

Remove usage of GraalVM

Fixes #852 

## Type of change

## Checklist:

## Additional notes:


Yauhenikapl (approved):
LGTM!

Yauhenikapl:
@e-filchenko-bosh just wanna ask you don't merge this PR before [Evaluate options to leverage SAST for GitHub action code](https://github.com/eclipse-esmf/esmf-sdk/pull/861), because here were fixed security issues in our workflows.

CC: @atextor

Yauhenikapl (changes requested):
Please, squash all totally same commits.

PR #934, opened 2026-04-14
fix: refactor AspectModelAsyncApiGenerator to use AspectModelJsonSchemaVisitor + minor fixes

### LINKED ISSUE #928: [BUG] Payload in generated AsyncAPI does not match specified in the model dataType
Open an aspect model with defined operation(s) and event(s).

Specified Operation is set as:
- name: getListOfItems
- input: itemId
- output: listOfItemsDetails where each list item dataType is: itemEntity

Specified Event is set as:
- name: updateItemDetails
- has two properties: itemId and itemtoUpdate, which dataType is: itemEntity

When generating AsyncAPI, it was expected that the created updateItemDetails operation would have a payload of two properties, as defined by the corresponding aspect model.
Instead, the updateItemDetails payload was set as String

This test was done with AME (v5.6.1 because latest 6.0.2 has some conflicts with our models, which we need to resolve). The AME v5.6.1 comes with ESMF SDK version: 2.7.0.

It was also tested with SAMM CLI v2.14.2 and had same result.

**Input (files/screenshots)**
**Real aspect model with Operation and Event**
Image

The below shown output is also similar, when I tried with simple aspect model, which has **only Event and Operation** and is referring the properties from "main" aspect model, so to avoid any properties duplication.

Image

**Operations in generated AsyncAPI**
Image

I've provided used aspect model(s) to @KobOp

andreas-wirth:
## Description
- refactor AspectModelAsyncApiGenerator to use AspectModelJsonSchemaVisitor to generate schema for asyncapi
- extend tests
- update asyncapi to 3.1.0

Fixes #928

## Type of change

## Checklist:

    No docs needed IMO

## Additional notes:


core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/asyncapi/AspectModelAsyncApiGenerator.java
```diff
@@ -98,176 +103,155 @@ public Stream<AsyncApiSchemaArtifact> generate() {
...
-         operation.getInput().forEach( input -> generateComponentsMessageAndSchemaOperation( messagesNode, schemasNode, input, locale ) );
+      aspect.getEvents().forEach( event ->
+            addOperation( operationsNode, aspect.getName(), event.getName(), ACTION_RECEIVE ) );
 
-         if ( operation.getOutput().isPresent() ) {
-            generateComponentsMessageAndSchemaOperation( messagesNode, schemasNode, operation.getOutput().get(), locale );
-         }
+      aspect.getOperations().forEach( operation -> {
+         operation.getInput().forEach( input ->
+               addOperation( operationsNode, aspect.getName(), input.getName(), ACTION_RECEIVE ) );
```
andreas-wirth, line 114: This generates an Operation for each input, which is not desirable. Instead, generate only one Operation with an object wrapping all inputs. Also consider adding suffixes to the generated messages like "<message>Request" and "<message>Response"

Kostadin-Ivanov: Thank you for noticing this, Andreas.

I added a question with example for this case, in related ticket, but now I see that you have mentioned it here.

Just to confirm - you are right that when we have an Operation with 1 or more inputs, all of these should be handled under 1 AsyncAPI Operation as inputs. 

Having multiple AsyncAPI operations for each individual input in the Aspect Model Operation **is wrong** because, in the model we would have defined specific Operation, which takes number of input parameters, that correspond to specific case or are required to be able to provide the declared in the aspect model operation response.
However, splitting the inputs in standalone AsyncAPI operations will break this set up, because it will provide # number of possible AsyncAPI calls that would be expected to return same response, which will be impossible due to the missing other input parameters.

With regards to the "Request" and "Response" - this would be very handy for better readability and understanding of the generated AsyncAPI, when one is going through it.

Thanks again and I hope that this would also be addressed in this fix :)

atextor (approved):

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/asyncapi/AspectModelAsyncApiGenerator.java
andreas-wirth: functionality is updated, checked with Kostadin in a separate chat.

atextor (approved):

Kostadin-Ivanov:
Hello Team and thank you for the great support on this :)

The AsyncAPI generator, now works as a charm :)

PR #974, opened 2026-06-16
Update AAS4J to 2.0.2

### LINKED ISSUE #973: Update AAS4J dependency to 2.0.2
ESMF currently still resolves AAS4J through `org.eclipse.esmf:esmf-parent:25`, which pins:

```xml
<aas4j.version>1.0.4</aas4j.version>
```

AAS4J `2.0.2` has been released and should be used by the SDK:
https://github.com/eclipse-aas4j/aas4j/releases/tag/2.0.2

This affects the AAS generator module dependencies:

- `aas4j-model`
- `aas4j-dataformat-xml`
- `aas4j-dataformat-aasx`
- `aas4j-dataformat-json`

While testing the update locally, two compatibility issues surfaced:

1. Generated AAS XML can contain invalid raw `idShort` values, for example a single-character property name like `a`. With AAS4J 2.0.2 this is serialized directly and fails the AAS XML schema pattern.
2. Imported IDTA AASX files can deserialize optional collections as `null`, for example `Entity.getStatements()` or `Reference.getKeys()`, which currently causes NPEs in the AAS-to-Aspect conversion path.

Expected outcome:
Update the SDK to AAS4J `2.0.2` and make the AAS generator/importer compatible with the stricter/raw serialization and nullable optional collections.

Acceptance criteria:
- AAS4J resolves to `2.0.2` for all AAS generator dependencies.
- Generated AAS XML remains schema-valid for the existing test aspect corpus.
- IDTA AASX test resources can still be translated without NPEs.
- `mvn -q -DskipITs test -pl core/esmf-aspect-model-aas-generator -am` passes.

stiebitzhofer:
## Summary

- Override the inherited AAS4J dependency version to `2.0.2`.
- Normalize generated AAS `idShort` values so AAS4J 2.0.2 raw XML serialization remains schema-valid.
- Handle nullable imported AAS collections such as `Entity.statements` and `Reference.keys` during AAS-to-Aspect conversion.

Fixes #973

## Verification

- `mvn -q spotless:apply -pl core/esmf-aspect-model-aas-generator -am`
- `mvn -q -DskipITs test -pl core/esmf-aspect-model-aas-generator -am`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
    <parent>
       <groupId>org.eclipse.esmf</groupId>
       <artifactId>esmf-parent</artifactId>
-      <version>26</version>
+      <version>27</version>
    </parent>
 
    <artifactId>esmf-sdk-parent</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
    <parent>
       <groupId>org.eclipse.esmf</groupId>
       <artifactId>esmf-parent</artifactId>
-      <version>25</version>
+      <version>26</version>
    </parent>
 
    <artifactId>esmf-sdk-parent</artifactId>
@@ -75,7 +75,6 @@
 
    <properties>
       <!-- General settings -->
-      <aas4j.version>2.0.2</aas4j.version>
       <skip.maven.surefire>false</skip.maven.surefire>
       <skip.maven.shade>false</skip.maven.shade>
       <skip.maven.failsafe>false</skip.maven.failsafe>
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,7 @@
 
    <properties>
       <!-- General settings -->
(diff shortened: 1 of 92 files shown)


pom.xml
```diff
@@ -75,6 +75,7 @@
 
    <properties>
       <!-- General settings -->
+      <aas4j.version>2.0.2</aas4j.version>
```
atextor: Please remove this line and instead update the esmf-parent at the top of the file from 25 to 26. [Version 26](https://github.com/eclipse-esmf/esmf-parent/releases/tag/v26) has been released to include aas4j 2.0.2.

atextor (changes requested):

pom.xml
stiebitzhofer: Sure, makes sense!

atextor (approved):
Looks good! Thank you for your contribution!

PR #976, opened 2026-06-16
Add AAS template qualifiers and example values

### LINKED ISSUE #435: [Task] CLI SAMM2AASX Add qualifier for cardinality

Up to now no qualifiers are added because SAMM defines these qualifiers like cardinality etc.
See https://industrialdigitaltwin.org/wp-content/uploads/2022/12/I40-IDTA-WS-Process-How-to-write-a-SMT-FINAL-.pdf Chapter 7.2

Nevertheless it might be useful to generate those as well to be consistent with the published other submodel templates.

For all optional elements add qualifier **Cardinality** with value **ZeroToOne**:

```
          <qualifiers>
            <qualifier>
              <semanticId>
                <type>ExternalReference</type>
                <keys>
                  <key>
                    <type>GlobalReference</type>
                    <value>https://admin-shell.io/SubmodelTemplates/Cardinality/1/0</value>
                  </key>
                </keys>
              </semanticId>
              <type>Cardinality</type>
              <valueType>xs:string</valueType>
              <value>ZeroToOne</value>
            </qualifier>
          </qualifiers>
```

### LINKED ISSUE #482: [BUG] Example values of properties are not used on AAS export
### Description
When an AAS export is performed, the example values of properties are not used at all.
The currently implemented logic is to use a payload value if a payload is given or the default (fallback) value if not.
(see implementation https://github.com/eclipse-esmf/esmf-sdk/blob/main/core/esmf-aspect-model-aas-generator/src/main/java/org/eclipse/esmf/aspectmodel/aas/Context.java#L178)

### Possible Solutions
The method `getPropertyValue` must take example values into account.
One option would be, to have an Optional chain trying to first get the payload value, then the example value, then the fallback value, e.g.

```java
public String getPropertyValue( final String defaultValue ) {
   return getRawPropertyValue().map( valueNode -> valueNode.asText( defaultValue ) )
         .or( () -> property.getExampleValue().map( ScalarValue::getValue ).map( Object::toString ) )
         .orElse( defaultValue );
}
```
(untested code, just for illustrational purposes)

This however could lead to a mix of payload and example and fallback values for optional payload values where the data is not given, however the model declares an example value.
If this is not desired, a second option could be to change the priorities in the chain and additionally skip example value retrieval if a payload is present, e.g.

```java
public String getPropertyValue( final String defaultValue ) {
   final Optional<String> exampleValue = aspectData == null ? property.getExampleValue().map( ScalarValue::getValue ).map( Object::toString ) : Optional.empty();
   return getRawPropertyValue().map( valueNode -> valueNode.asText( defaultValue ) )
         .or( () -> exampleValue )
         .orElse( defaultValue );
}
```
(untested code, just for illustrational purposes)

### LINKED ISSUE #797: [Task] CLI SAMM2AASX - geneate example values
generate example values if available in Aspect Model

- as value
- as template qualifier "SMT/ExampleValue"

stiebitzhofer:
## Description

This PR enriches generated AAS templates with cardinality and example-value metadata:

- replaces the optional-only cardinality qualifier with template-only cardinality qualifiers for scalar and collection properties
- maps SAMM example values into generated template values when no instance payload is present
- adds `SMT/ExampleValue` qualifiers for properties with SAMM example values
- keeps actual payload data preferred over example values for instance generation

Fixes #435
Fixes #482
Fixes #797

## Validation

```sh
mvn -q -DskipITs test -pl core/esmf-aspect-model-aas-generator -am
mvn -q spotless:apply -pl core/esmf-aspect-model-aas-generator -am
```

Both completed successfully locally.


atextor (approved):
Looks good to me. Thanks for your contribution!

PR #992, opened 2026-06-29
[Task] Implement cross file definition service for SAMM LSP

### LINKED ISSUE #990: [Task] Support cross file "Go To Definition" for SAMM LSP
SAMM LSP should support to go definitions across different files.

- re-use filesystemstrategy to resolve local files
- Also support resolution of samm internal definitions

A clear and concise description of any alternative solutions or features you have considered.

andreas-wirth:
## Description

- Implement "Go To Definition" across workspace and SAMM elements for SAMM LSP
- Fix behavior for `isSammUrn` check in AspectModelUrn
- Fix validation to properly load aspect model dependencies
- Introduce async execution of lsp requests

Fixes #990

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
+++ b/core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
@@ -13,7 +13,7 @@
 
 package org.eclipse.esmf.treesitterturtle;
 
-import java.util.Arrays;
+import java.nio.charset.StandardCharsets;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
@@ -224,7 +224,15 @@ public class TurtleSyntaxTree {
          return new Error( inputNode.getType(), TurtleDiagnostic.TurtleCode.E0004, location, children );
       }
       final Supplier<String> token = () -> new String(
-            Arrays.copyOfRange( content.getBytes(), inputNode.getStartByte(), inputNode.getEndByte() ) );
+            // bytes
+            content.getBytes( StandardCharsets.UTF_8 ),
+            // offset
+            inputNode.getStartByte(),
+            // length
+            inputNode.getEndByte() - inputNode.getStartByte(),
+            // charset
+            StandardCharsets.UTF_8
+      );
       return new Token( inputNode.getType(), token, location, children );
    }
 
--- a/core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
+++ b/core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.esmf.treesitterturtle;
 
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
@@ -32,8 +33,8 @@ import org.treesitter.TSTree;
  * Represents the concrete syntax tree of a turtle document
  */
 public class TurtleSyntaxTree {
-   private final Supplier<String> sourceRepresentationSupplier;
    private final Node rootNode;
+   private final String content;
 
    public static class TurtleSyntaxTreeTraversalError extends RuntimeException {
    }
@@ -80,13 +81,18 @@ public class TurtleSyntaxTree {
     */
    public record Token(
          String type,
-         String content,
+         Supplier<String> contentSupplier,
          Location location,
          List<Node> children
    ) implements Node {
       @Override
       public Token asToken() {
          return this;
+      }
+
+      @Override
+      public String content() {
+         return contentSupplier.get();
       }
    }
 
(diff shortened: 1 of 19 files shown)


atextor (changes requested):
Hi, the implementation looks good so far, only a couple of notes:
* The meta model definition files opened when jumping to the definition of SAMM elements happen to be considered Aspect Models (via `ParsedDocument#isAspectModel`). This is true only for the files associated with an `ElementType` other than "meta-model". So, by clicking e.g. `samm:description`, the temp file for aspect-meta-model-definitions.ttl opens, and this file should _not_ be subject to Aspect Model validation.
* Jumping to definitions from the `unit` namespace doesn't seem to work, e.g. to `unit:metre`.
* After trying the latter (jump to `unit:metre`'s def), the definition service seems to "hang", and jumping doesn't work any more until a client restart

core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
```diff
@@ -211,14 +216,15 @@ private static Node nodeForTsNode( final TSNode inputNode, final TokenProvider t
...
+            .map( child -> nodeForTsNode( child, content ) )
             .toList();
       if ( inputNode.isError() ) {
          return new Error( inputNode.getType(), TurtleDiagnostic.TurtleCode.E0003, location, children );
       } else if ( inputNode.isMissing() ) {
          return new Error( inputNode.getType(), TurtleDiagnostic.TurtleCode.E0004, location, children );
       }
-      final String token = tokenProvider.apply( location );
+      final Supplier<String> token = () -> new String(
+            Arrays.copyOfRange( content.getBytes(), inputNode.getStartByte(), inputNode.getEndByte() ) );
```
atextor: Instead of copying the token subsection of the array, we should construct the String by directly referencing the subsection (content is immutable anyways):

```java
final Supplier<String> token = () -> new String( 
  // bytes
  content.getBytes( StandardCharsets.UTF_8 ),
  // offset
  inputNode.getStartByte(),
  // length
  inputNode.getEndByte() - inputNode.getStartByte(),
  // charset
  StandardCharSets.UTF_8
);
```

That way, we don't duplicate memory usage for large documents.

core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
```diff
@@ -80,14 +81,19 @@ default Error asError() {
...
    ) implements Node {
       @Override
       public Token asToken() {
          return this;
       }
+
+      @Override
+      public String content() {
+         return contentSupplier.get();
+      }
```
atextor, line 96: What is the benefit of this change? In particular, when adjusting the construction of token Strings (see below), I think this change can be reverted.

core/esmf-turtle-language-server/src/main/java/org/eclipse/esmf/turtle/languageserver/lsp/LspUtil.java
```diff
@@ -0,0 +1,37 @@
...
+
+   private LspUtil() {
+   }
+
+   public static ResolutionStrategy buildResolutionStrategyForDocument( final ParsedDocument parsedDocument ) {
+      final Path openFilePath;
+      try {
+         openFilePath = Path.of( new URI( parsedDocument.getUri() ) );
+      } catch ( URISyntaxException e ) {
+         throw new RuntimeException( e );
```
atextor: A ResolutionStrategy should never throw RuntimeException, but ModelResolutionException instead, ideally using the constructor taking a collection of checked locations (for better error reporting).

Also, this method should probably check that the URI is indeed a `file:` URI (which is not necessarily the case) and throws a corresponding exception if it isn't.

core/esmf-turtle-language-server/src/test/java/org/eclipse/esmf/turtle/languageserver/turtle/navigation/TurtleCrossFileDefinitionServiceTest.java
```diff
@@ -157,33 +151,40 @@ void testReturnsEmptyWhenElementNotFound( @TempDir( cleanup = CleanupMode.NEVER
...
 
-      final Optional<Location> result = resolver.findDefinition( parsedDocument, new Position( 4, 19 ) );
+      final Optional<Location> result = resolver.findDefinition( parsedDocument, new Position( 8, 20 ) );
 
-      assertThat( result ).isEmpty();
+      assertThat( result ).isPresent();
+      final Location location = result.get();
+      assertThat( location.getUri() ).contains( "unit" );
+      assertThat( location.getUri() ).endsWith( ".ttl" );
+      assertThat( location.getRange().getStart() ).isEqualTo( new Position( 11712, 5 ) );
```
atextor: Maybe rather check for some location >1000 or so, because otherwise the check will fail if the specific formatting/location of the file changes in the future.

atextor (changes requested):

core/esmf-tree-sitter-turtle/src/main/java/org/eclipse/esmf/treesitterturtle/TurtleSyntaxTree.java
atextor: I see the benefit; ohne for those Tokens for which the content() is acutally inspected, corresponding strings are actually constructed.

atextor (approved):
LGTM!

PR #999, opened 2026-07-07
Change the subsystem flag in the Windows executable from GUI to CUI

### LINKED ISSUE #997: [BUG] Windows binary of samm-cli launches another console when running

On Windows:

```
PS C:\Users\sekik> Invoke-WebRequest https://github.com/eclipse-esmf/esmf-sdk/releases/download/v2.15.0/samm-cli-2.15.0-windows-x86_64.zip -OutFile samm-cli-2.15.0-windows-x86_64.zip
PS C:\Users\sekik> Expand-Archive .\samm-cli-2.15.0-windows-x86_64.zip
PS C:\Users\sekik> .\samm-cli-2.15.0-windows-x86_64\samm.exe --version
```

Actual result: the command above launches another console (see the image below), which closes so quickly after completing the task that users can't see the results.

Expected result: it should run within the same window.

* [samm-cli-2.15.0-windows-x86_64.zip](https://github.com/eclipse-esmf/esmf-sdk/releases/download/v2.15.0/samm-cli-2.15.0-windows-x86_64.zip)
* Windows 11 Pro 25H2

**Input (files/screenshots)**

Image

The current binary contains a flag that marks it as a GUI application.

```
PS C:\Users\sekik> python3 -c "import pefile; print(pefile.PE('samm-cli-2.15.0-windows-x86_64\\samm.exe').OPTIONAL_HEADER.Subsystem)"
2
```

[According to the Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#windows-subsystem), the value `2` indicates a GUI application. Changing this value to `3` (`IMAGE_SUBSYSTEM_WINDOWS_CUI`) seems to work. Or, the `--hidden` option [newly added in warp v1.3.0](https://github.com/kirbylink/warp/releases/tag/v1.3.0) may work as well.

sekikn:
## Description

This PR changes the subsystem flag in the Windows executable so that it behaves as a CUI application and doesn't launch another console window.

Fixes #997 

## Type of change

## Checklist:

  - No explicit comments, as the file and function names are self-explanatory enough.

  - No corresponding documentation

- [ ] I have added tests that prove my fix is effective or that my feature works

  - No additional tests are included because the fix is for the release process. Instead, I ran the same workflow (except for disabling the builds for Linux and macOS and the release job) within my own GitHub repository and ensured the output binary worked as expected.

Screenshot from 2026-07-07 22-19-14

a


atextor:
Hi @sekikn, thank you so much for this contribution!
I've pushed a commit that also incorporates the fix into the PR build, so that it allows me to check the binary that is built there.

atextor:
I can confirm that the fix works as expected for the generated .exe; in that context I'd approve and merge the PR. However, there's one problem left related to the release process of the Windows version of samm-cli: The samm-cli Windows .exe needs to be signed using the Eclipse Foundation certificate, which is only possible on the Eclipse CI infrastructure but not GitHub Actions.

The process works as follows:
1. The release-workflow.yml is executed: Java libraries and samm-cli Linux, Windows and Mac binaries are built
2. GitHub release is created, containing Linux and Mac binaries (but not Windows)
3. Java libraries are uploaded to Maven Central
4. Windows binary is uploaded to the build's cached artifacts
5. A remote Jenkins pipeline running on Eclipse CI is triggered via an API call. The pipeline itself is configured in [.jenkins/Jenkinsfile](https://github.com/eclipse-esmf/esmf-sdk/blob/main/.jenkins/Jenkinsfile).
6. The Jenkins pipeline fetches the Windows binary from the build's cached artifacts, signs it and uploads it to the corresponding GitHub release

Steps 5+6 are configured here: [release-workflow.yml#L389-L427](https://github.com/eclipse-esmf/esmf-sdk/blob/main/.github/workflows/release-workflow.yml#L389-L427).

So for the approach to work for releases, setting the flag would need to be done in step 6 in the Jenkins-Pipeline, after signing and before uploading it to the GH release.

sekikn:
Thank you for the elaboration @atextor, but I still can't understand the statement below, because modifying binary content (like changing a flag) **after** signing it is supposed to cause verification error, I think.

> So for the approach to work for releases, setting the flag would need to be done in step 6 in the Jenkins-Pipeline, after signing and before uploading it to the GH release.

Here's an example. At the point of downloading the v2.15.0 binary for Windows, that signature is valid as the following screenshot (sorry that the text is Japanese, but you can see the normal icon at the top-left).

image

But after changing the subsystem flag, that signature is no longer valid, as follows (you can see the error icon in the screenshot).

```
$ ipython
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.15.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can use LaTeX or Unicode completion, `\alpha<tab>` will insert the α symbol.

In [1]: import pefile

In [2]: pe = pefile.PE("samm.exe")

In [3]: pe.OPTIONAL_HEADER.Subsystem
Out[3]: 2

In [4]: pe.OPTIONAL_HEADER.Subsystem = 3

In [5]: pe.write("samm.exe")

In [6]: pe.close()
```

image

But I'm not confident if I properly understand the release and signing process of Eclipse Foundation, so correct me if my understanding is wrong.

atextor (approved):
You are right. I had the wrong mental model of "flag" in my mind: It's not a file system flag like "executable" which needs to be set on the newly created binary, but it's actually a change inside the binary itself, therefore it needs to be done before signing.

PR #1001, opened 2026-07-08
Generate valid annotation

### LINKED ISSUE #978: [BUG] DTO generator doesn't place @Valid for classes within Optional<List<Object>>
### Summary

When generating Java DTOs from a `.ttl` model, the DTO generator correctly adds Bean Validation annotations for entity types inside plain lists, but it does not do the same when the `List` is wrapped in an `Optional`.

For an optional property whose characteristic is a `samm-c:List` of an entity type, the generated DTO currently contains:

```java
@Valid
private Optional<List<MetadataEntity>> metadata;
```

However, the expected output should validate the list elements as well:

```java
@Valid
private Optional<List<@Valid MetadataEntity>> metadata;
```

The generator already appears to do this correctly for non-optional lists:

```java
@Valid
private List<@Valid MetadataEntity> metadata;
```

So the issue seems specific to the combination of `samm:optional true` plus `samm-c:List` whose `samm:dataType` is an entity.

### Minimal reproduction

Given the following `.ttl` model:

```ttl
@prefix samm: <urn:samm:org.eclipse.esmf.samm:meta-model:2.1.0#> .
@prefix samm-c: <urn:samm:org.eclipse.esmf.samm:characteristic:2.1.0#> .
@prefix : <urn:samm:io.catenax.requirement:1.0.0#> .

:RequirementInformationEntity a samm:Entity ;
   samm:description "Entity to describe all information for a requirement"@en ;
   samm:properties ( [ samm:property :metadata; samm:optional true ] ) .

:metadata a samm:Property ;
   samm:preferredName "Metadata"@en ;
   samm:description "Additional metadata that is associated with the requirement and not standardized in the aspect model."@en ;
   samm:characteristic :MetadataList .

:MetadataList a samm-c:List ;
   samm:preferredName "Metadata List"@en ;
   samm:description "List of additional metadata that is associated with the requirement and not standardized in the aspect model."@en ;
   samm:dataType :MetadataEntity .

:MetadataEntity a samm:Entity ;
   samm:preferredName "Metadata Entity"@en ;
   samm:description "Entity for key and value of requirement metadata."@en ;
   samm:properties ( :key :value [ samm:property :metadataDescription; samm:optional true ] ) .

:metadataDescription a samm:Property ;
   samm:preferredName "Metadata Description"@en ;
   samm:description "Description of the associated metadata"@en ;
   samm:characteristic samm-c:Text ;
   samm:exampleValue "Timestamp of the expected finalization of the requirement" .
```

### Actual generated DTO field

```java
@Valid
private Optional<List<MetadataEntity>> metadata;
```

### Expected generated DTO field

```java
@Valid
private Optional<List<@Valid MetadataEntity>> metadata;
```

### Why this matters

With the current generated code, validation is applied to the `Optional` wrapper, but not to the individual `MetadataEntity` instances inside the nested `List`.

As a result, constraints inside `MetadataEntity` may not be triggered when the property is optional but present.

For consistency with non-optional list generation, and to ensure nested entity validation works correctly, the generated type should include `@Valid` on the list element type:

```java
Optional<List<@Valid MetadataEntity>>
```

### Expected behavior

For an optional list of entity objects, the DTO generator should generate Bean Validation annotations equivalent to:

```java
@Valid
private Optional<List<@Valid MetadataEntity>> metadata;
```

More generally, when a property is both optional and a list whose item type is an entity, the generator should preserve the same element-level validation behavior that it already applies to non-optional lists.

### Actual behavior

The generator emits:

```java
@Valid
private Optional<List<MetadataEntity>> metadata;
```

This loses the nested `@Valid` annotation on the list element type.

### Suggested fix direction

The DTO generator likely already has logic for producing:

```java
List<@Valid SomeEntity>
```

for non-optional list properties. The same logic should probably be applied when the list type is wrapped in `Optional`, producing:

```java
Optional<List<@Valid SomeEntity>>
```

instead of:

```java
Optional<List<SomeEntity>>
```

### Environment

Used esmf-sdk 2.14.1

e-filchenko-bosh:
## Description

Fixes Java DTO generation for collection properties whose element type is a SAMM Entity. The generator now adds Bean Validation `@Valid` as a type-use annotation on collection element types, so optional wrappers preserve element validation, e.g. `Optional<Collection<@Valid TestEntity>>`.

Existing collection element constraints from `elementCharacteristic` / `Trait` remain supported.

No new dependencies are required.

Fixes #978

## Type of change

## Checklist:

### DIFF
--- a/core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaUtil.java
+++ b/core/esmf-aspect-model-java-generator/src/main/java/org/eclipse/esmf/aspectmodel/java/AspectModelJavaUtil.java
@@ -58,6 +58,7 @@ import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.google.common.base.CaseFormat;
 import com.google.common.base.Converter;
+import jakarta.validation.Valid;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
 import org.apache.commons.text.translate.UnicodeUnescaper;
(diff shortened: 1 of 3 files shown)


andreas-wirth (approved):

atextor (approved):
LGTM!

PR #1003, opened 2026-07-16
AsyncAPI generator: tenant-id in channel address, expanded channel parameters, and fixed Operation input rendering

### LINKED ISSUE #942: [Task] Make AsyncAPI generation consistent with OpenAPI spec generation (url path includes <tenant-id>)
The OpenAPI spec generated url includes a tenant-id (by default, configurable), see https://eclipse-esmf.github.io/esmf-developer-guide/tooling-guide/samm-cli.html#using-the-cli-to-create-a-json-openapi-specification .

This may also be desirable for the AsyncAPI channel address (see https://eclipse-esmf.github.io/esmf-developer-guide/tooling-guide/samm-cli.html#using-the-cli-to-create-a-json-asyncapi-specification ) in order to foster use cases for digital twins that require such a path element.

The solution should be backwards compatible, i.e. allows defining channels without tenant-id.
The solution should make OpenAPI and asyncAPI behavior as consistent with each other as possible while following the conventions and typical usability of AsyncAPI

Currently, as a workaround, the ca channel address parameter can be used.

**Additional context**

gjavakhadze:
### Problems described in #942

1. **Tenant-id in channel address**: The AsyncAPI generator has no way to include a tenant-id in the generated channel address, unlike the OpenAPI generator, which already supports this for its URLs.
2. **Channel parameter expansion**: Channel parameters (`namespace`, `version`, `aspect-name`, `tenant-id`) are generated as plain static strings instead of full AsyncAPI Parameter Objects.
3. **Operation input rendering**: Multiple Operation Inputs are generated as an `allOf` array of anonymous schemas, which the AsyncAPI Editor renders as an "Anonymous Schema" and only displays the last input in the example payload.

This PR fixes all three problems.

---

### Tenant-id and channel parameter expansion

- The default channel address is now `/{tenant-id}/{namespace}/{version}/{aspect-name}`, fully templated instead of interpolating literal values.
- Each of the `tenant-id`, `namespace`, `version`, `aspect-name` channel parameters is rendered as a full AsyncAPI Parameter Object (`description`, `default`) only when its corresponding placeholder is literally present in the channel address — whether that's the default address or an explicit `channelAddress` override — so the `parameters` section always matches what the address actually uses.

These were initially introduced as opt-in behaviors behind an `AsyncApiGenerationFeature` flag, but are now made the generator's permanent, unconditional behavior — the flag/config option has been removed.

### Operation input rendering

Operation inputs are now represented as named properties on a single `object` schema instead of an `allOf` array of anonymous schemas. Each input is keyed by its payload name under `properties`, which fixes tooling (e.g. the AsyncAPI Editor) rendering the combined input as an unlabeled "Anonymous Schema" and only showing the last input in generated examples.

Fixes #942

## Type of change

## Checklist:


core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/asyncapi/AsyncApiSchemaGenerationConfig.java
```diff
@@ -28,17 +29,27 @@
...
+ *        set (the default) leaves generator output unchanged.
  */
 @RecordBuilder
 public record AsyncApiSchemaGenerationConfig(
       Locale locale,
       boolean useSemanticVersion,
       String applicationId,
-      String channelAddress
+      String channelAddress,
+      Set<AsyncApiGenerationFeature> features
```
atextor: Introducing `AsyncApiGenerationFeature` mixes the approaches for representing config settings (or features if you will). While I think both are valid, we should only have one of them:
1. The current approach: A `record` where each field is one configuration flag. By using `@RecordBuilder`, order of fields for constructing objects is not important, because you can always use the generated builder. The class `AsyncApiSchemaGenerationConfigBuilder` is automatically generated at build time from the record by the record-builder library (and annnotation processor). E.g.:
      ```java
      var config = AsyncApiSchemaGenerationConfigBuilder.builder()
         .locale( Locale.ENGLISH )
         .useSemanticVersion( true )
         /* ...other fields as required */
         .build();
      ```

2. Feature flags as enum as introduced in `AsyncApiGenerationFeature`.

Since the pattern of 1. is already widely used, I suggest adapting it, by removing `AsyncApiGenerationFeature` and instead using regular fields in the record:


```suggestion
      String channelAddress,
      boolean addTenantIdInChannelParameters,
      boolean expandChannelParameters
```

Instead of
```java
if ( config.hasFeature( AsyncApiGenerationFeature.ADD_TENANT_ID_IN_CHANNEL_PARAMETERS ) )
```
you'd be calling
```java
if ( config.addTenantIdInChannelParameters() )
```

Note: If you adjust the record and want to have the corresponding generated builder updated, you need to re-run compilation with the record builder agent, i.e., `mvn -pl core/esmf-aspect-model-document-generators clean compile`.

atextor (changes requested):

core/esmf-aspect-model-document-generators/src/main/java/org/eclipse/esmf/aspectmodel/generator/asyncapi/AsyncApiSchemaGenerationConfig.java
gjavakhadze: I completely removed flags, so, there is no more newly introduced concepts in code

core/esmf-aspect-model-document-generators/src/test/java/org/eclipse/esmf/aspectmodel/generator/asyncapi/AspectModelAsyncApiGeneratorTest.java
```diff
@@ -426,4 +469,126 @@ void testAsyncApiGeneratorAspectWithEventAndEntityProperty() throws IOException
...
+      final JsonNode json = asyncSpec.getContent();
+
+      final JsonNode expectedChannels = OBJECT_MAPPER.readTree(
+            """
+                  {
+                     "Aspect": {
+                        "address": "123/456/test/1.0.0/TestAspect",
+                        "description": "Channel for updating Aspect Aspect.",
+                        "parameters": {
+                           "namespace": {
```
andreas-wirth: The channel address does not include parameters, but we still list them here. Is this really desirable? Maybe we want to default to only listing parameters that are actually used. WDYT?

gjavakhadze: Your observation seems logical to me, but in this change I only expanded parameters section and did not introduce change their logic when to appear.

Screenshot 2026-07-22 at 16 22 32

BTW I can make parameters to be a dynamic according to the channel address. That is how newly added `tenantId` parameter is managed.

andreas-wirth: From my point of view it would be desirable to have consistent behavior. So if the tenant-id disappears if not used, the rest should also disappear if not used in the address

gjavakhadze: @andreas-wirth Good point. I think to make all async generator channel parameters dynamic and parameterized. Default address value will be: `/{tenant-id}/{namespace}/{version}/{aspect-name}`

In case user overrides channel address, parameters section will contain only mentioned parameters. So channel address `/{tenant-id}/fleet-generic-vehicle-telemetry-async-api` will contain only tenant-id parameter only.

andreas-wirth (approved):

documentation/developer-guide/modules/tooling-guide/pages/samm-cli.adoc
```diff
@@ -1042,7 +967,14 @@ samm aspect _Movement.ttl_ to asyncapi -ff ExpandChannelParameters
 }
 ----
 
-Both flags can be combined, e.g. `-ff AddTenantIdInChannelParameters,ExpandChannelParameters`.
+Each of the `tenant-id`, `namespace`, `version` and `aspect-name` channel parameters is rendered as a full
+https://www.asyncapi.com/docs/reference/specification/v3.1.0#parameterObject[Parameter Object] only when the
+corresponding placeholder (`\{tenant-id}`, `{namespace}`, `{version}`, `{aspect-name}`) is literally present in the
+channel address, so the `parameters` section always matches what the address actually uses. If no explicit channel
+address is configured via `--channel-address`/`-ca` (Maven: `channelAddress`), the default address
```
atextor: The information about the Maven plugin switches should go to maven-plugin.adoc.

atextor (changes requested):

documentation/developer-guide/modules/tooling-guide/pages/samm-cli.adoc
gjavakhadze: @atextor Updated the PR according to your note

atextor (approved):
Looks good to me!

PR #1004, opened 2026-07-17
LSP: Fix diagnostics referring to other files

atextor:
## Description

Aspect Model violations that accur for problems in resolved files are currently reported for the originally validated file. This PR:

1. Introduces the abstraction of a "DiagnosticsProvider" for model validation - any
service that takes as input a ParsedDocument (RDF/Turtle source document with
its syntax tree) and provides validations about it.

Parts that are specific to SAMM Aspect Models are refactored into corresponding
packages. Turtle analysis (e.g. syntax checks) are separated from Aspect Model
validation. Common functionality, such as the ValidationCoordinator (running
delayed validation for slower DiagnosticsProviders) is made generic, since it's
not specific to Aspect Models.

2. Used the new structure to correctly report Aspect Model validations as diagnostics that refer to other files using `DiagnosticRelatedInformation`.

Fixes https://github.com/eclipse-esmf/esmf-vs-code-plugin/issues/14.

## Type of change

## Checklist:

## Additional notes:

In VS Code, the problems view displays as a nested problem with a link to the referred file.


andreas-wirth (approved):
Especially like the idea to configure fast / delayed validation on interface level 👍

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/constraint/JsConstraint.java
```diff
@@ -122,33 +124,32 @@ public List<Violation> apply( final RDFNode rdfNode, final EvaluationContext con
...
-               new IllegalArgumentException() ) );
+            default -> {
+               LOG.debug( "JavaScript evaluation of {} returned invalid result: {}", jsFunctionName(), result );
+               yield List.of( new ProcessingViolation( "JavaScript evaluation of " + jsFunctionName() + " returned an invalid result",
+                     new IllegalArgumentException() ) );
+            }
+         };
       } catch ( final ScriptException exception ) {
-         LOG.debug( "JavaScript evaluation of {} failed", jsFunctionName(), exception );
-         return List.of( new ProcessingViolation( "JavaScript evaluation of " + jsFunctionName() + " failed", exception ) );
```
e-filchenko-bosh, line 149: I think this change may hide real validation problems.

Before this PR, `ScriptException` and `NoSuchMethodException` were reported as `ProcessingViolation`. Now both cases return an empty list, so a broken JS constraint or a missing JS function can make the model look valid even when the check was not executed.

This feels risky because `JsConstraint` is part of the general validator, not only the language server. Maybe we should keep returning `ProcessingViolation` here and handle the noisy/duplicate diagnostic closer to the LSP layer, or filter only the specific known case

core/esmf-turtle-language-server/src/main/java/org/eclipse/esmf/turtle/languageserver/lsp/diagnostic/DiagnosticMapper.java
```diff
@@ -0,0 +1,80 @@
...
+   public Map<URI, List<Diagnostic>> apply( final Document sourceDocument, final DiagnosticReport report ) {
+      final Map<URI, List<Diagnostic>> result = new HashMap<>();
+      for ( final org.eclipse.esmf.Diagnostic<?> lspDiagnostic : report.diagnostics() ) {
+         final Diagnostic diagnostic = new Diagnostic();
+         diagnostic.setSeverity( toDiagnosticSeverity( lspDiagnostic.severity() ) );
+         diagnostic.setMessage( lspDiagnostic.message() );
+         diagnostic.setCode( lspDiagnostic.code().code() );
+         if ( lspDiagnostic instanceof final DocumentDiagnostic<?> documentDiagnostic ) {
+            if ( documentDiagnostic.sourceLocation().equals( sourceDocument.uri() ) ) {
+               diagnostic.setRange( toRange( documentDiagnostic ) );
```
e-filchenko-bosh, line 48: I think this improves the situation, but it may still be a bit misleading for cross-file diagnostics.

When `documentDiagnostic.sourceLocation()` is different from `sourceDocument.uri()`, we add `DiagnosticRelatedInformation` that points to the real location. However, the main diagnostic is still published for `sourceDocument.uri()` in the first line.

So if validation is triggered for `A.ttl`, but the real problem is in `B.ttl`, the Problems view will still show the diagnostic on `A.ttl`. The user can navigate to `B.ttl` through related information, but the primary error is not attached to the file where the problem actually is.

Is it expected behaviorus?

core/esmf-turtle-language-server/src/main/java/org/eclipse/esmf/turtle/languageserver/lsp/diagnostic/DiagnosticMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+   public Map<URI, List<Diagnostic>> apply( final Document sourceDocument, final DiagnosticReport report ) {
+      final Map<URI, List<Diagnostic>> result = new HashMap<>();
+      for ( final org.eclipse.esmf.Diagnostic<?> lspDiagnostic : report.diagnostics() ) {
+         final Diagnostic diagnostic = new Diagnostic();
+         diagnostic.setSeverity( toDiagnosticSeverity( lspDiagnostic.severity() ) );
+         diagnostic.setMessage( lspDiagnostic.message() );
+         diagnostic.setCode( lspDiagnostic.code().code() );
+         if ( lspDiagnostic instanceof final DocumentDiagnostic<?> documentDiagnostic ) {
+            if ( documentDiagnostic.sourceLocation().equals( sourceDocument.uri() ) ) {
```
e-filchenko-bosh, line 47: Could we add a regression test for the main cross-file diagnostic case?

The new `DiagnosticMapper` logic adds `DiagnosticRelatedInformation` when the diagnostic source location is different from the currently validated document. This is the key behavior for the issue, but I do not see a test that covers `documentDiagnostic.sourceLocation() != sourceDocument.uri()`.

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/constraint/JsConstraint.java
atextor: You are correct that broken script validation is not producing ProcessingViolations any more, and this intentional. Failure to produce violations should be fixed on our side, but it's not due to something that the Aspect Model author did wrong, so it should not be reported in that channel.

core/esmf-turtle-language-server/src/main/java/org/eclipse/esmf/turtle/languageserver/lsp/diagnostic/DiagnosticMapper.java
atextor: Yes, you are completely correct in your assessment. And yes, it's intentional after trying out several variations. In the beginning, I implemented it like this:
* I have `A.ttl` open and trigger validation, which finds a violation in `B.ttl`
* The diagnostic is only reported for `B.ttl`. However, it is not automatically opened in VSCode (and it probably shouldn't be). So validation failed for my currently edited file, but I can't see where or why.

So in the current implementation, the problem is reported for both `A.ttl` and `B.ttl`, but the one on `A.ttl` has the attached information: "B.ttl: Root cause is here" and you can double-click it and it takes you to the other file. Without the reported diagnostic for `A.ttl`, you couldn't do that.

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/shacl/constraint/JsConstraint.java
e-filchenko-bosh: Hm, I fully agree that it should not be reported in that channel, but currently we cannot identify the problem and we cannot receive a report by user, because we just hide this error. This also can affect CLI validation and it can be critical.

Should we change something here, for example by logging it at **warn** level with the exception?

atextor: I think the evaluation of JSConstraints in the validation called from the LSP should be fixed and the fix checked using a regular test. That way the corresponding failure is not reported to the user (false negative) but we still can be sure that it works as expected. I've created a corresponding issue: https://github.com/eclipse-esmf/esmf-sdk/issues/1006

e-filchenko-bosh (approved):

PR #1025, opened 2026-08-07
Add fix for scalar value for array in json-example generation

### LINKED ISSUE #1024: [BUG] .json generation error in Lists
## Summary

When a Property uses a Collection Characteristic (`samm-c:Set`/`List`) but only carries a single `exampleValue`, `to json` writes that value as a bare scalar instead of wrapping it in a one-element array — even though `to schema` correctly generates `"type": "array"` for that same property. So the example payload and the schema, generated from the same model by the same tool, contradict each other.

I tested this with multiple released CLI versions. It seems the issue appeared since version 2.13.0 and above (worked perfectly with previous versions). You can find a minimum example below; or a concrete model: https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.pcf/10.0.0/Pcf.ttl. However, the issue appears for all models with the same constellation below.

1. Model has a `Property` whose `Characteristic` is a `samm-c:Set` (Collection type), with a single `samm:exampleValue` literal set directly on the `Property` (not an RDF-list of multiple values). Example from a real model:
   ```turtle
   :companyIds a samm:Property ;
      samm:characteristic :IdSet ;
      samm:exampleValue "urn:BPNL000000000DWF"^^xsd:anyURI .

   :IdSet a samm-c:Set ;
      samm:dataType xsd:anyURI .
   ```
2. Generate the example JSON payload: and JSON Schema
4. Validate `payload.json` against `schema.json`
5. Repeated across samm-cli versions 2.11.1 through 2.15.1 to bisect when the regression was introduced.

Minimal reproduction snippet (isolated from the real model, for convenience):
```turtle
@prefix samm: <urn:samm:org.eclipse.esmf.samm:meta-model:2.1.0#> .
@prefix samm-c: <urn:samm:org.eclipse.esmf.samm:characteristic:2.1.0#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix : <urn:samm:com.example:1.0.0#> .

:MinimalRepro a samm:Aspect ;
   samm:properties ( :ids ) ;
   samm:operations ( ) ;
   samm:events ( ) .

:ids a samm:Property ;
   samm:characteristic :IdSet ;
   samm:exampleValue "urn:example:1234"^^xsd:anyURI .

:IdSet a samm-c:Set ;
   samm:dataType xsd:anyURI .
```

---

*Expected:* since the Characteristic is a `samm-c:Set`, the generated JSON Schema correctly declares the property as `"type": "array"`. The generated example payload should therefore wrap the single `exampleValue` in a one-element array, e.g. `"companyIds": ["urn:BPNL000000000DWF"]`, matching its own schema.

*Actual:* from **2.13.0 onward**, `to json` emits the bare scalar instead: `"companyIds": "urn:BPNL000000000DWF"`. This does not validate against the schema samm-cli generates for the very same property/model:
```
'urn:BPNL000000000DWF' is not of type 'array'
```
The schema side (`to schema`) is correct and unaffected — only the example-payload generator (`to json`) regressed.

---

- **First bad version: 2.13.0.** Last known-good: 2.12.0.
- Confirmed present in: 2.13.0, 2.14.3, 2.15.0, 2.15.1
- Confirmed absent in: 2.11.1, 2.12.0
- Affected real-world model: [io.catenax.pcf v10.0.0](https://github.com/eclipse-tractusx/sldt-semantic-models/blob/main/io.catenax.pcf/10.0.0/Pcf.ttl), property `companyIds` (line 409, Characteristic `:IdSet` at line 736) — also reproduces on `productIds` and `productClassifications` in the same model, so it affects every `samm-c:Set`/`List`-typed property with a single `exampleValue`, not just this one.

- Potentially related to the previously-fixed issue with the same symptom: #423 ("Example json Payload generated as plain value but should be array"), fixed in 2023, appears reintroduced somewhere in the 2.12.0 → 2.13.0 range.

jSchuetz88:
Note in advance: I'm not an Expert in Java and had to work a little with AI to understand the Code. I used the codebase from previous versions of the CLI (2.12.0) as an example, as the issue from #1024 was not present in this version. I tried to follow all style guides and to introduce a minimum invasive change.

This PR resolved the issue I faced in #1024, but should be reviewed carefully.

Closes #1024

Edit: I might should have used the 2.15.x branch as the base?

## Description

`to json` emitted a bare scalar value for a Property whose Characteristic is a Collection (`samm-c:Set`/`List`/`SortedSet`) when the Property carried only a single scalar `exampleValue`, instead of wrapping it in a one-element JSON array. This contradicted `to schema`, which correctly declares `"type": "array"` for the same property, so generated example payloads failed schema validation.

The regression was introduced in the `JsonPayloadGenerator` rewrite in 2.13.0 (visitor pattern), which bypassed the Collection-aware visit methods when an `exampleValue` was present.

Fix: `visitProperty()` now checks `property.getEffectiveCharacteristic()` (also unwraps `samm-c:Trait`) and wraps the value in a one-element array when the effective Characteristic is a Collection.

Fixes #(number of issue in GitHub)

## Type of change

## Checklist:

## Additional notes:

- Style verified with `mvn spotless:check` and `mvn checkstyle:check` on the changed module — both pass.
- Self-review included a line-by-line comparison against the last known-good pre-2.13.0 implementation (commit `92070b6d`) to confirm behavioral parity for the reported scenario.
- No documentation changes: this restores previously-documented/intended behavior, no user-facing behavior or API changed beyond what was already documented.
- Added `testGenerateJsonForAspectWithSetCharacteristicAndSingleExampleValue` and `testGenerateJsonForAspectWithListCharacteristicAndSingleExampleValue`; both run and pass under `mvn test`. Also verified end-to-end against the PCF:10.0.0 example referenced in #1024 using a locally built `samm-cli` jar.


atextor (approved):
Hi @jSchuetz88, thanks for your contribution!
You indeed identified a regression, and the fix looks good to me.

For future contributions, you generally can branch from `main`. The release branches, such as `2.15.x`, are intended to allow for backporting fixes to older minor versions (which rarely happens).

PR #1026, opened 2026-08-10
Update openapi generator

### LINKED ISSUE #1016: [Task] make special behavior for HTTP code 200 when generating OpenAPI spec using a template
**Description**
Version [esmf-sdk v2.15.1](https://github.com/eclipse-esmf/esmf-sdk/releases/tag/v2.15.1) improves OpenAPI spec generation using a template. Now, when using a template with `__DEFAULT_QUERIES_TEMPLATE__ `placeholder, only those HTTP codes in the response section that are specified in the template are generated.
```
paths:
  __DEFAULT_QUERIES_TEMPLATE__:
    get:
      responses:
        '400':
          $ref: "core-api.yaml#/components/responses/ClientError"
        '404':
          $ref: "core-api.yaml#/components/responses/NotFound"
```
So, now user must specify the 200 HTTP response code as well in the template, otherwise generated OpenAPI spec will not contains a point for this code. However, each aspect model has its own model for the 200 HTTP response (e.g., `$ref: "#/components/responses/BatteryInformationPublic"`, `$ref: "#/components/responses/Plant"`, `$ref: "#/components/responses/Batch"` etc). Therefore, it is not possible to specify the 200 HTTP response code in a common template using the `__DEFAULT_QUERIES_TEMPLATE__` placeholder.

**The solution I'd like:**
My proposal is to add an exception for the 200 HTTP response code and return it to its previous behavior: if the 200 HTTP code is explicitly specified in the template, it will be used to generate the OpenAPI spec. If it's not specified, a default value should be generated based on the aspect model.
 
**Actual behavior:**
Template contains only 400 and 404 HTTP codes, generated OpenAPI spec contains only these codes.
 
**Expected behavior:**
Template contains only 400 and 404 HTTP codes, but generated OpenAPI spec contains these codes + 200 HTTP code based on aspect model.

e-filchenko-bosh:
## Description

Previously, defining `responses` in the operation template replaced the whole generated response section, including the model-specific generated success response. This made reusable error response templates impractical, because the generated `200` response depends on the Aspect model or operations endpoint.

With this change:
- generated error responses are replaced when custom template `responses` are provided;
- generated success response is preserved by default;
- template-defined `responses."200"` overrides the generated success response;
- `x-esmf-generate-success-response: false` disables adding the generated success response

Fixes #1016

## Type of change

## Checklist:


atextor (approved):
LGTM!

PR #1027, opened 2026-08-11
Harmonize diagnostic and violation

### LINKED ISSUE #1029: [Task] De-duplicate behaviour between Violation and Diagnostic
In the code base, there are multiple representions of "something that went wrong": The `Violation` interface - which is intended to represent SHACL engine validation results, hence the method to return the SHACL engine `EvaluationContext` - and the `Diagnostic` interface in the LSP. This is error-prone and confusing; handling and processing is done in multiple places in multiple ways.

Refactor so that only one interface (e.g. `Violation`) is left. Extract SHACL-specifics into a separate interface. Adjust LSP and ViolationFormatters. Move interface into esmf-aspect-model-interfaces module. Simplify validator interfaces where possible (remove overly complicated generic type signatures).

atextor:
## Description

This PR streamlines and cleans up validation reporting. It includes the following high-level changes:

* What was previously known as `Violation` was split into `Violation` (the generic part) and `ShaclViolation` (which includes the SHACL engine `ExecutionContext`). The new `Violation` interface was then moved to the esmf-aspect-meta-model-interface module. Lastly, the `Diagnostic` interface from the LSP was merged into the new Violation interface, which therefore becomes the single top-level type for all kinds of violations/validations. 
* A new explicit type instead of `List<Violation>` was introduced, the `ViolationReport`.
* A clean type hierarchy for various kinds of Violations is introduced (only most important types listed here):
  * `Violation` is the top level interface. `ProcessingViolation` and `ModelResolutionViolation` (previously known as `ModelResolutionException.LoadingFailure`) are the only direct implementations.
  * `DocumentViolation` is the subtype that knows about the file the violation was raised in
  * `DocumentLocationViolation` is the subtype of the above that also knows about line/column
  * `ElementFocussedViolation` is the subtype of the above that also knows about the RDFNode the violation refers to
  * `ShaclViolation` is a subtype of the above that also knows about the SHACL engine violation state
* With Violation being present in the interfaces module, there is no need any more for the type arguments in the validator interfaces. For example `AspectModelBasedValidator<P, C extends Collection<? super P>>` is now only `AspectModelBasedValidator`. This change affects both AspectModelValidator and AspectModelLoader. Instead of `C`, methods now just return `ViolationReport`.
* Since the new `Violation` is supposed to be extensible in the future, it's not a closed hieararchy and does not have a Visitor interface any more (`ShaclViolation` keeps its Visitor, though). `ViolationFormatter` and `DetailedViolationFormatter` are changed accordingly (they don't implement the Visitor interface; multiplexing violation rendering logic is adjusted).
* Since the Violation type hierarchy requires certain fields to be present (e.g., the document URI in `DocumentViolation`), the optionality for source locations is removed: The source URI of a document is now systematically tracked throughout the parsing/loading/validation cycle. This affects `TurtleLoader`, `TokenRegistry`, `RawAspectModelFile` and several other places that dealt with the `Optional<URI>`.
* In the LSP, the `AspectViolationDiagnosticMapper` is removed completely, since no mapping is necessary any more.
* In the LSP, the `AspectModelValidationService` has been adjusted to only use `AspectModelValidator`'s resulting `ViolationReport`, it does not copy the exception dispatching logic any more.

Fixes #1029

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

## Additional notes:


tools/esmf-aspect-model-maven-plugin/src/test/java/org/eclipse/esmf/aspectmodel/ValidateTest.java
```diff
@@ -54,7 +54,7 @@ public void testValidateValidAspectModel( final Validate validate ) {
    public void testValidateInvalidAspectModel( final Validate validate ) {
       assertThatCode( validate::execute )
             .isInstanceOf( MojoExecutionException.class )
-            .hasMessageContaining( "Error at line 17 column 3" );
+            .hasMessageContaining( "Error at line" );
```
andreas-wirth: Why does the message not contain the line and column anymore?

andreas-wirth (approved):

core/esmf-aspect-model-validator/src/main/java/org/eclipse/esmf/aspectmodel/validation/services/AspectModelValidator.java
```diff
@@ -96,50 +105,103 @@ public List<Violation> validateModel( final Supplier<AspectModel> aspectModelSup
...
+         return Either.left( new ViolationReport( exception.problems().stream()
+               .<Violation>map( problem -> new MetaModelVersionViolation( problem.message(), problem.highlight() ) )
+               .toList() ) );
       } catch ( final CancelValidation cancelValidation ) {
          // The validation was short-circuited by the aspectModelLoader function
          return Either.left( cancelValidation.violations );
+      } catch ( final AspectLoadingException exception ) {
+         return Either.left( reportForAspectLoadingException( exception ) );
+      } catch ( final ModelResolutionException exception ) {
+         return Either.left( reportForModelResolutionException( exception ) );
```
e-filchenko-bosh, line 134: I think this branch accidentally drops some ModelResolutionExceptions.
reportForModelResolutionException() only maps exception.getCheckedLocations() into the report. But ModelResolutionException can also be created with just a message/cause, in which case checkedLocations is empty. Before this change, those cases were handled by the generic catch (Exception) and became a ProcessingViolation. Now they are caught earlier by the dedicated catch (ModelResolutionException) and become an empty ViolationReport, so validation/LSP may treat a real loading failure as “no findings”.
If checkedLocations is empty, we should fall back to the default exception handling and create a ProcessingViolation from the exception message/cause

e-filchenko-bosh (approved):

tools/esmf-aspect-model-maven-plugin/src/test/java/org/eclipse/esmf/aspectmodel/ValidateTest.java
atextor: Good point, I removed it temporarily because the numbers didn't match due to off-by-one. I'm reintroducing them.

